PR #6, opened 2018-11-28
Feature/multisource

dpastoor:


dpastoor:
multisource installations are working, tested across osx and ubuntu

PR #26, opened 2018-12-11
fix: error if non-200 result returns with html instead of PACKAGES

### LINKED ISSUE #25: crash when querying PACKAGES file in invalid URL
In the case that the server does not return a 404, rather some HTML that notes that the query is invalid when querying a packages file, the parsing then fails.

```
problem parsing package with info  <!DOCTYPE html>
<html>
  <head>
    <meta http-equiv="Content-type" content="text/html; charset=utf-8">
    <meta http-equiv="Content-Security-Policy" content="default-src 'none'; style-src 'unsafe-inline'; img-src data:; connect-src 'self'">
    <title>Page not found &middot; GitHub Pages</title>
    <style type="text/css" media="screen">
      body {
        background-color: #f1f1f1;
        margin: 0;
        font-family: "Helvetica Neue", Helvetica, Arial, sans-serif;
      }
Bad line: '<!DOCTYPE html>
' has no ':'
panic: Bad line: '<!DOCTYPE html>
' has no ':'

goroutine 4 [running]:
github.com/metrumresearchgroup/pkgr/cran.(*RepoDb).FetchPackages.func1(0xc000078040, 0xc00002e0c0, 0x0, 0x0, 0x2)
	/Users/devinp/go/src/github.com/metrumresearchgroup/pkgr/cran/repodb.go:154 +0xdae
created by github.com/metrumresearchgroup/pkgr/cran.(*RepoDb).FetchPackages
	/Users/devinp/go/src/github.com/metrumresearchgroup/pkgr/cran/repodb.go:105 +0x518
```

This can arise when querying for a sourcetype that doesn't exist, such as binaries when only source is available and the webserver returns some html body

dpastoor:
also catch goroutine panic on closed channel as if the first repo failed
the func would return and close the channel before the
other goroutine could send

fixes #25

### DIFF
--- a/cmd/plan.go
+++ b/cmd/plan.go
@@ -15,6 +15,7 @@
 package cmd
 
 import (
+	"strings"
 	"fmt"
 	"os"
 	"time"
@@ -64,7 +65,16 @@ func planInstall() (*cran.PkgDb, gpsr.InstallPlan) {
 		}
 	}
 	st := cran.DefaultType()
-	cdb, err := cran.NewPkgDb(repos, st, cran.NewPkgConfigDB())
+	cic := cran.NewInstallConfig()
+	for rn, val := range cfg.Customizations.Repos {
+		if strings.EqualFold(val.Type, "binary") {
+			cic.Repos[rn] = cran.RepoConfig{DefaultSourceType: cran.Binary}
+		}
+		if strings.EqualFold(val.Type, "source") {
+			cic.Repos[rn] = cran.RepoConfig{DefaultSourceType: cran.Source}
+		}
+	}
+	cdb, err := cran.NewPkgDb(repos, st, cic)
 	if err != nil {
 		fmt.Println("error getting pkgdb ", err)
 		panic(err)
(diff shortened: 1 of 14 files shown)


dpastoor:
digging into this also demonstrated a number of problematic areas around how packages in repos for various source types are reconciled, namely on a binary platform, there wasn't a clean way to express "install only source packages from this repo", be it for user facing reasons, or because binary version(s) did not exist.

This refactoring and addition changes the config file as well, but will support a better abstraction around how to customize a repo.

For example, in the following, all packages in r_validated will pull source versions, which is key as no binary versions exist for osx/windows in that particular repository.

```
...
Repos:
  - r_validated: "https://metrumresearchgroup.github.io/r_validated"
  - CRAN: "https://cran.rstudio.com"

Library: "testsets/multirepo"

Cache: "testsets/pkgcache"

Logging:
  File: "multirepo.log"

# customizations are package specific settings
# for example, we can use a custom makevar file for data.table
# by leveraging the R_MAKEVARS_USER to point to a separate makevar
# each package gets installed in its own process so any env variables
# or settings will only matter for it
Customizations:
  Repos:
      - r_validated:
          Type: source  
```

PR #32, opened 2019-01-20
Feat/env

dpastoor:
add the capability to set environment variables during installation of packages. 

The overall internal representation is to now store two sets of env variables - global ones that should be applied to every package during install, as well as local ones per package.


dpastoor:
completed with tests added around environment configuration at a package level

Issue #36, opened 2019-02-10
cache not recognized since changing to unique repo naming


dpastoor:
relevant line of code

```
	bpath := filepath.Join(pc.BaseDir, meta.Metadata.Config.Repo.Name, "binary", binaryName(pkg.Package, pkg.Version))
```

however now the package is downloaded and installed to the directory based on the URL hash

```
urlHash := RepoURLHash(d.Config.Repo)
pkgdir := filepath.Join(baseDir, urlHash, pkgType)
```

dpastoor:
tested by manually deleting a random package (haven) and re-installing with logging to trace levels

Relevant records

```
TRAC[0001] found in cache                                package=haven path=/Users/devinpastoor/rpkgs/pkgcache/CRAN-ae21d9639812/binary/3.5/haven_2.1.0.tgz
TRAC[0001] command args                                  RSettings="{{3 5 2} []  map[] map[]}" cmd=install cmdArgs="[--vanilla CMD INSTALL --install-tests --no-multiarch --with-keep.source --library=/Users/devinpastoor/rpkgs/3.5 /Users/devinpastoor/rpkgs/pkgcache/CRAN-ae21d9639812/binary/3.5/haven_2.1.0.tgz]" env="REMOVED" package=haven
DEBU[0002] cmd output                                    exitCode=0 package=haven stderr="* installing *binary* package ‘haven’ ...\n* DONE (haven)\n" stdout=
DEBU[0002] package install request completed             WID=10 duration=364.448338ms package=haven
DEBU[0002] haven                                         binary= src=/Users/devinpastoor/rpkgs/pkgcache/CRAN-ae21d9639812/binary/3.5/haven_2.1.0.tgz
INFO[0002] Successfully Installed                        package=haven repo=CRAN version=2.1.0
```

PR #45, opened 2019-02-25
Refactor/caching

dpastoor:


dpastoor:
overall tested by running a fresh install of packages on OSX.

* inspected `Cache` for proper 
* investigated removing packages will install next from cache #36 

One thing to note, is if a package is set to install from source, it will re-compile from source no matter if a binary version is available in the cache. 

```
DEBU[0003] cmd output                                    exitCode=0 package=devtools stderr="* installing *source* package ‘devtools’ ...\n** package ‘devtools’ successfully unpacked and MD5 sums checked\n** R\n** inst\n** tests\n** byte-compile and prepare package for lazy loading\n** help\n*** installing help indices\n*** copying figures\n** building package indices\n** installing vignettes\n** testing if installed package can be loaded\n* creating tarball\npackaged installation of ‘devtools’ as ‘devtools_2.0.1.tgz’\n* DONE (devtools)\n" stdout=
TRAC[0003] binary location prior to install              bbp=devtools_2.0.1.tgz binaryBall=/var/folders/g_/0948z6350756j3yly26_zqnr0000gn/T/devtools_2.0.1.tgz tbp=/Users/devinpastoor/rpkgs/pkgcache/CRAN-ae21d9639812/src/devtools_2.0.1.tar.gz
TRAC[0003] command args                                  RSettings="{{3 5 2} [/Users/devinpastoor/rpkgs/3.5]  map[] map[]}" cmd=install cmdArgs="[--vanilla CMD INSTALL --library=/Users/devinpastoor/rpkgs/3.5 /var/folders/g_/0948z6350756j3yly26_zqnr0000gn/T/devtools_2.0.1.tgz]" env="REMOVED" package=devtools
DEBU[0004] cmd output                                    exitCode=0 package=devtools stderr="* installing *binary* package ‘devtools’ ...\n* DONE (devtools)\n" stdout=
DEBU[0004] package install request completed             WID=6 duration=3.604029585s package=devtools
DEBU[0004] devtools                                      binary=/var/folders/g_/0948z6350756j3yly26_zqnr0000gn/T/devtools_2.0.1.tgz src=/Users/devinpastoor/rpkgs/pkgcache/CRAN-ae21d9639812/src/devtools_2.0.1.tar.gz
```

I think this (for now) is a reasonable trade-off, as setting `Type=source` could very well be because you want explicit compiler flags or otherwise. That said, one common use on OSX is if the src version is ahead of the binary, and so you want to compile from source particularly to get the newer version.

Thus, this might be better to have something like `Recompile=true/false` with a default value of false, so if a binary version of that is available, use it. Hence the flow would be:

```
* check source version
* if binary version of source version is available + recompile=false, use binary
* if binary version of source not available or recompile=true, compile from source
```

Issue #49, opened 2019-03-17
clean/invalidate pkgdb


dpastoor:
similar to #40.

Case: new package has been published to cran-like repos in rpkgs (or otherwise). Would like to install, however have run pkgr recently, so the pkgdb cache still contains the old db for at least 1 hr. 

## scoping expectations:

* Need an api proposal (eg how would users do this) 
* potential code areas of note
* estimated time
* risk assessment

dpastoor:
api implemented and discussed in #40

Issue #40, opened 2019-02-16
clean/invalidate caches


dpastoor:
should clean at least:

- [x] package cache
  - [x] binary
  - [x] src

package cache located at `Cache`

## scoping expectations:

* Need an api proposal (eg how would users do this) 
* potential code areas of note
* estimated time
* risk assessment

Dreznel:
## General thoughts and relevant code sections

After a pass through some parts of the code, I this seems to just require clearing a directory/items from a directory that is set by this function call: https://github.com/metrumresearchgroup/pkgr/blob/aca7e894242db6b2f18c3886152f857f0d565b7c/cmd/install.go#L57.

That call includes a call to the userCache function from utils.go in the same package.

## API Proposal
The easiest way to let users do this would probably be to just add a "pkgr clean" subcommand that both clears the cache and the pkgdb. If we don't want this "clear all" by default, we could add sub-subcommands (sort of like git does) such as "pkgr clean cache", "pkgr clean database <database, "" = -A for all>", and "pkgr clean -A" for all.

I'm proposing a general solution here because I like the idea of a "pkgr clean -A" or "pkgr reset --hard" command to fully reset the state of your pkgr setup, but in reality I think that #49 should probably have a more nuanced solution that checks for updates.

## Estimated Time
To just add the "clear the cache folder" subcommand would probably take me between four and twelve hours to implement, and between four and eight hours to write tests for (I'll have to learn how to use afero, which I'm unfamiliar with). These are rough estimates, obviously, but that's what it feels like assuming my assessments are correct.

Dreznel:
We could also pass in the "clear cache" option as a flag on the `pkgr install` command, but I think that this would result in a more complicated change.

dpastoor:
Great proposal, lets do the following:

* all

```
pkgr clean --all
```

* crandbs

```
pkgr clean --pkgdb    // all
pkgr clean --pkgdb="CRAN"    // specific db
pkgr clean --pkgdb="CRAN,r_validated" // comma separated slice
pkgr clean --pkgdb="CRAN, r_validated" // whitespace should be ok?
```

* package cache
   * binary
    * src

```
pkgr clean --cache
pkgr clean --cache // CRAN/binary or CRAN/src, etc ...? 
pkgr clean --cache --src= --binary= ?
```

PR #52, opened 2019-03-21
Feature/invalidate caches

Dreznel:
Working on #40 and #49

### DIFF
--- a/cmd/cleanCache.go
+++ b/cmd/cleanCache.go
@@ -15,16 +15,19 @@
 package cmd
 
 import (
-	"fmt"
+	"errors"
 	"path/filepath"
 	"strings"
 
+	"github.com/sirupsen/logrus"
+	log "github.com/sirupsen/logrus"
 	"github.com/spf13/cobra"
 )
 
-var cleanCache bool
-var srcCaches string
-var binaryCaches string
+//var cleanAllFromCache bool
+var srcOnly bool
+var binariesOnly bool
+var reposToClear string
 
 // cacheCmd represents the cache command
 var cleanCacheCmd = &cobra.Command{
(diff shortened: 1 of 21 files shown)


integration_tests/internal/pkgr.yml
```diff
@@ -22,14 +22,13 @@ Suggests: true
 Repos:
   - r_validated: "https://metrumresearchgroup.github.io/r_validated"
   - CRAN: "https://cran.rstudio.com"
-  - local: "path/on/filesystem"
```
dpastoor: if we're going to test local, it needs to be a relative path to the pkgr file so it can also be generalized across machines

integration_tests/internal/pkgr.yml
```diff
@@ -22,14 +22,13 @@ Suggests: true
...
+  - local: "/Users/johncarlos/openspace/rpkgs/gh_dev"
 
-Library: "path/to/installation"
+Library: "test-library"
 
 # any libpaths desired to be set
 Libpaths:
-  - "path/to/some/prev/installed"
-  - "another/path/to/some/prev/installed"
+  - "testsets/simple"
```
dpastoor: shouldn't probably need any extra libpaths, this is only if there are existing packages ont he system you want included in the package set

integration_tests/internal/pkgr.yml
```diff
@@ -45,3 +44,5 @@ Customizations:
              - R_MAKEVARS_USER: "~/.R/Makevars_data.table"
         - internal:
             Repo: local
+        - mrgsolve:
```
dpastoor: this is adding extra complexity to this 'test' which is to test local install of an internal package, should be removed/pulled to a separate test dir

integration_tests/internal/pkgr.yml
Dreznel: Right, I haven't finished cleaning this up but I figured I'd move the ones I messed with into their own folders for now.

integration_tests/internal/pkgr.yml
Dreznel: So is it safe to remove these from everything?

dpastoor: yes - we'll want to test that at some point but overall its somewhat of an antipattern

eg you should really have all your deps in one place, but this is an escape hatch in case you need it

cmd/cleanCache.go
```diff
@@ -0,0 +1,106 @@
...
+	if err != nil {
+		return err
+	}
+
+	repoFolders, _ := cacheDirFsObject.Readdir(0)
+
+	if repos == nil || len(repos) == 0 {
+		for _, repoFolder := range repoFolders {
+			subfolderPath := filepath.Join(cacheDirectory, repoFolder.Name(), subfolder)
+			fs.RemoveAll(subfolderPath)
```
dpastoor: don't forget to check if that worked with an err := fs.RemoveAll()

cmd/cleanCache.go
```diff
@@ -0,0 +1,106 @@
...
+		}
+	} else {
+		for _, repoToClear := range repos {
+			for _, repoFolder := range repoFolders {
+				if repoToClear == repoFolder.Name() {
+					subfolderPath := filepath.Join(cacheDirectory, repoFolder.Name(), subfolder)
+					fs.Remove(subfolderPath)
```
dpastoor: Why remove vs removeall

Dreznel: Good question.

Issue #18, opened 2018-12-09
logging option to delete existing log file


dpastoor:
especially helpful if running commands successively in a testing context. 

Default should remain to keep the same file, as it provides a comprehensive audit log.

## considerations

* [ ] decouple file log separately for installation vs planning
	* plans do not need to log to the log file by default

Dreznel:
## Proposed Acceptance Criteria:

* Logging options can be set in the config file to do one of the following
    * **default**: append to the existing log file without ever deleting it
    * delete existing log whenever a pkgr command is executed
* Logging option is created to control whether logs all go to one file, or whether pkgr install gets its own log file.
    * **default**: all pkgr commands go to a single log file.
    * optionally, the user can use the config file to indicate a separate log file, which is then used exclusively for logs from pkgr install commands.
* Code is modified to prevent pkgr plan from logging to file.
    * logging option is created to override this behavior and allow pkgr plan to log to file.

dpastoor:
yes these look great - how do you plan on setting the api in the config file?

PR #58, opened 2019-03-27
Feature/logging config

Dreznel:
Setting this pull request base as the branch I branched off of. We'll have to either merge this into that branch when it's done, or set up a new pull request to go straight into the feature branch.

This pull request addresses #18

### DIFF
--- a/cmd/plan.go
+++ b/cmd/plan.go
@@ -25,8 +25,9 @@ import (
 	"github.com/metrumresearchgroup/pkgr/configlib"
 	"github.com/metrumresearchgroup/pkgr/cran"
 	"github.com/metrumresearchgroup/pkgr/gpsr"
+	Log "github.com/metrumresearchgroup/pkgr/logger"
 	"github.com/sajari/fuzzy"
-	log "github.com/sirupsen/logrus"
+	"github.com/sirupsen/logrus"
 	"github.com/spf13/cobra"
 	"github.com/spf13/viper"
 )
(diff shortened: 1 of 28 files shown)


Dreznel:
A note:

Logrus achieves its exported Log by basically rewriting wrapper functions for every function on the Logger type, each of which calls the corresponding instance method on the "std" log. In order not to do this ourselves, I propose we follow the pattern of importing from out own local logger package as ".", then just calling log statements with the "Log.[...]" (capital L) syntax. This is very similar to how it's working now.

cmd/plan.go
```diff
@@ -118,7 +119,7 @@ func planInstall(rv cran.RVersion) (*cran.PkgDb, gpsr.InstallPlan) {
 			if configlib.IsCustomizationSet("Type", pkgSettings, pkg) {
 				err := cdb.SetPackageType(pkg, v.Type)
 				if err != nil {
-					log.WithFields(log.Fields{
+					log.WithFields(logrus.Fields{
```
dpastoor: hmm there is some inconsistencies with your naming, above I saw you refactoring logrus.Fields to log.Fields (fine), but there going in the opposite direction. I'd pick one and stick to it

Dreznel: So that was intentional -- I thought that, since the logger package is the only one really dealing directly with logrus, I would keep the full name there to avoid confusion. I can still change it if you want, but I prefer it this way.

Dreznel: Ooh, nevermind, just took another look. I'll loook into this

logger/logger.go
```diff
@@ -3,8 +3,9 @@ package logger
 import (
 	"fmt"
 	"os"
+	"strings"
 
-	"github.com/sirupsen/logrus"
+	logrus "github.com/sirupsen/logrus"
```
dpastoor: for consistency lets call this `log` as well

logger/logger.go
```diff
@@ -14,6 +15,17 @@ type LogrusFileHook struct {
 	formatter *logrus.JSONFormatter
 }
 
+// Log Reinstantiable logrus to be used globally in the application.
+//var Log *logrus.Logger
+
+func init() {
+	//Log = logrus.New()
+	logrus.SetOutput(os.Stdout)
+	logrus.SetFormatter(&logrus.TextFormatter{ForceColors: true})
```
dpastoor: update to `log`

PR #63, opened 2019-04-02
Feature/version detection

Dreznel:
We don't have a release 3.0 branch yet, so I'm just creating a merge request to 2.0 so that we can start reviewing.


cmd/plan.go
```diff
@@ -162,7 +169,72 @@ func planInstall(rv cran.RVersion) (*cran.PkgDb, gpsr.InstallPlan) {
...
+}
+
+func GetPriorInstalledPackages(fileSystem afero.Fs, libraryPath string) map[string]InstalledPackage {
+
+	installed := make(map[string]InstalledPackage)
+
+	installedLibrary, err := fileSystem.Open(libraryPath)
+
+	if err != nil {
+		//panic?
```
dpastoor: probably stop with an informative error message ("no path detected at...")

cmd/plan.go
```diff
@@ -162,7 +169,72 @@ func planInstall(rv cran.RVersion) (*cran.PkgDb, gpsr.InstallPlan) {
...
+		return installed
+	}
+
+	installedPkgs, _ := installedLibrary.Readdir(0)
+
+	for _, f := range installedPkgs {
+		descriptionFilePath := filepath.Join(libraryPath, f.Name(), "DESCRIPTION")
```
dpastoor: this whole chunk of code should be broken out to a separate set of function(s). For example readDescription (returns a byte slice or string slice). 

getPackageMetadata() PkgMetadata 

where the PkgMetadata struct would have like Version/Name/other?

these could then be tested and managed independently and probably should be in a separate package entirely

cmd/plan.go
```diff
@@ -162,7 +169,72 @@ func planInstall(rv cran.RVersion) (*cran.PkgDb, gpsr.InstallPlan) {
...
+		descriptionFile, err := fileSystem.Open(descriptionFilePath)//, _ := fs.Open()
+
+		if err != nil {
+			//panic?
+			log.WithField("file", descriptionFilePath).Warn("DESCRIPTION missing from installed package")
+		}
+		scanner := bufio.NewScanner(descriptionFile)//#bufio.Scanner(fs.Open(f))
+		log.WithField("description file", descriptionFilePath).Debug("scanning DESCRIPTION file")
+
+		var pkgName, pkgVersion, pkgRepo = "", "", ""
```
dpastoor: empty string is the default value for these so can just initialize as type string

cmd/plan.go
```diff
@@ -162,7 +169,72 @@ func planInstall(rv cran.RVersion) (*cran.PkgDb, gpsr.InstallPlan) {
...
+			splitLine := strings.Split(scanner.Text(), ":")
+			switch strings.ToLower(splitLine[0]) {
+			case "package":
+				pkgName = strings.TrimSpace(splitLine[1])
+			case "version":
+				pkgVersion = strings.TrimSpace(splitLine[1])
+			case "repository":
+				pkgRepo = strings.TrimSpace(splitLine[1])
+			default:
+				log.WithField("line", scanner.Text()).Debug("no info found on line")
```
dpastoor: there will be a lot of additional info found so this debug log will be huge (namely all non pkg/version/repo lines in every description file)

cmd/plan_test.go
```diff
@@ -0,0 +1,116 @@
...
+	"fmt"
+	"github.com/dpastoor/goutils"
+	"github.com/spf13/afero"
+	//"os"
+	"github.com/stretchr/testify/assert"
+	"path/filepath"
+	"testing"
+)
+
+func InitializeTestEnvironment(goldenSet, testName string) afero.Fs{
```
dpastoor: check the suite https://github.com/stretchr/testify subpackage of testify, which I think is does a lot of these things for you.

cmd/plan_test.go
```diff
@@ -0,0 +1,116 @@
...
+
+	libraryPath, _ := filepath.Abs(filepath.Join("testsite", "working", "basic-test1", "test-library"))
+
+	actual := GetPriorInstalledPackages(fileSystem, libraryPath)
+
+	assert.Equal(t, 2, len(actual))
+	assert.True(t, installedPackagesAreEqual(expectedResult1, actual[expectedResult1.Name]))
+	assert.True(t, installedPackagesAreEqual(expectedResult2, actual[expectedResult2.Name]))
+
+	DestroyTestEnvironment()
```
dpastoor: what happens if an assert fails? will the testenvironment ever get cleaned up. If you're going to manually handle destruction, you want to make sure to do so such that it will almost certainly run (eg barring program crashing).

Namely, as soon as you perform the activity that would clean up, you should `defer` the destruction, such that it will definitely get run when the function exists.

So on line 74 it might basically be

```
InitializeTestEnvironment()
defer DestroyTestEnvironment()
```

cmd/testsite/golden/basic-test1/test-library/R6/INDEX
```diff
@@ -0,0 +1,3 @@
+R6Class                 Create an R6 reference object generator
```
dpastoor: Do we really care about checking in all information from the package, especially the INDEX and binary RDS files? Wouldn't we just want to keep around the folder/DESCRIPTION file?

cmd/plan_test.go
Dreznel: I'll need to leave a good chunk of InitializeTestEnvironment as it is, but I have started to implement the setup and teardown functions, and that will definitely be helpful for future testing efforts.

Dreznel:
including #69 in this PR as of commit 3680d43

Dreznel:
@dpastoor 
So I can still think of a few ways to clean this up, but for now I've got a working first-pass at #69. 

My approach so far is:
1. Attach outdated packages object to the InstallPlan
2. IFF the "update" flag is set, move all outdated packages into backup directories ("tagged with __OLD__" before installation begins. This causes the install command to install those packages regularly. Save all "attempted updates" into an object we can reference later.
3. At the end of the installation process, scan all of the attempted updates and make sure they completed. We verify that an updated succeed by making sure there is a folder in the library with the update package's name. If any failures are found, restore from saved backups. Either way, remove the saved backups.

cmd/utils.go
```diff
@@ -42,3 +49,212 @@ func getWorkerCount() int {
...
+	}
+
+	return updateAttempts
+}
+
+func tagOldInstallation(fileSystem afero.Fs, libraryPath string, outdatedPackage gpsr.OutdatedPackage) UpdateAttempt {
+	packageDir := filepath.Join(libraryPath, outdatedPackage.Package)
+	taggedPackageDir := filepath.Join(libraryPath, "__OLD__" + outdatedPackage.Package)
+
+	err := RenameDirRecursive(fileSystem, packageDir, taggedPackageDir)
```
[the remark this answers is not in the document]
Dreznel: fileSystem.Rename() doesn't work properly -- files and directories within the renamed directory do not have their paths altered. The unit tests fail if I implement it that way.

Basically we need  a way to do the Rename recursively, so at that point is was easiest just to copy everything recursively.

cmd/utils.go
```diff
@@ -42,3 +49,212 @@ func getWorkerCount() int {
 	}
 	return nworkers
 }
+
+func GetPriorInstalledPackages(fileSystem afero.Fs, libraryPath string) map[string]desc.Desc {
+	installed := make(map[string]desc.Desc)
+	installedLibrary, err := fileSystem.Open(libraryPath)
+
+	if err != nil {
+		log.WithField("libraryPath", libraryPath).Fatal("package library not found at given library path")
```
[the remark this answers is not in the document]
Dreznel: I believe you are correct, but I decided it was best to treat the code as if it would NOT crash, since log levels might be changed. In general, I don't like relying on a logging object to crash the program, but I guess I can see the logic there.

The important thing here is that control flow breaks -- I don't want someone in the future to change the log-level to "Error" and inadvertently cause the function to do weird things.

Thoughts?

cmd/utils.go
```diff
@@ -42,3 +49,212 @@ func getWorkerCount() int {
...
+
+	descriptionFile, err := fileSystem.Open(descriptionFilePath)
+
+	if err != nil {
+		log.WithField("file", descriptionFilePath).Warn("DESCRIPTION missing from installed package.")
+		return desc.Desc{}, err
+	}
+	defer descriptionFile.Close()
+
+	log.WithField("description file", descriptionFilePath).Debug("scanning DESCRIPTION file")
```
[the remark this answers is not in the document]
Dreznel: Agreed, good catch.

cmd/utils.go
```diff
@@ -42,3 +49,212 @@ func getWorkerCount() int {
...
+	if installedPackage.Package != "" {
+		return installedPackage, nil
+	} else {
+		err = errors.New(fmt.Sprintf("encountered a description file without package name: %s", descriptionFile))
+		log.WithField("description file", descriptionFilePath).Error(err)
+		return desc.Desc{}, err
+	}
+}
+
+func GetOutdatedPackages(installed map[string]desc.Desc, availablePackages cran.AvailablePkgs) []gpsr.OutdatedPackage {
```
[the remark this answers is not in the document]
Dreznel: Good catch! Will fix. 

All I have to do is add an if statement that checks if installedVersion != availableVersion.

cmd/utils.go
```diff
@@ -42,3 +49,212 @@ func getWorkerCount() int {
...
+	err = fileSystem.RemoveAll(oldPath)
+	if err != nil {
+		return err
+	}
+
+	return nil
+}
+
+//TODO: Move into goutils.
+func CopyDir(fs afero.Fs, src string, dst string) error {
```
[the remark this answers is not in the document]
Dreznel: Agreed. Ultimately, I want to add this to your goutils package that we're using elsewhere. This was originally only for setting up tests, but because of the weirdness with fs.Rename(), I ended up promoting this to a real function. It needs cleaning, but I figure we'll do that (and add tests) when we transfer it between repos.

cmd/utils.go
```diff
@@ -42,3 +49,212 @@ func getWorkerCount() int {
...
+
+	installedPackageFolders, _ := installedLibrary.Readdir(0)
+
+	for _, pkgFolder := range installedPackageFolders {
+		descriptionFilePath := filepath.Join(libraryPath, pkgFolder.Name(), "DESCRIPTION")
+		installedPackage, err := scanInstalledPackage(descriptionFilePath, fileSystem)
+
+		if err != nil {
+			log.Error(err)
+			err = nil
```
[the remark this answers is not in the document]
Dreznel: This takes place in a for loop, so it's possible that err might persist between loop runs (since it's technically defined outside of the loop)

cmd/utils.go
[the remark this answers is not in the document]
dpastoor: but the entire intent of Fatal is to explicitly cause everything to exit. If you _don't_ want that behavior then you drop to an error and return. 

You're intent:

> I believe you are correct, but I decided it was best to treat the code as if it would NOT crash, since log levels might be changed

actually is more confusing, as when reading the code, you say hmm Fatal should be exiting, but then there is a return after so the code is now telling me 2 different behavioral stories. Which is the expected behavior?

cmd/utils.go
```diff
@@ -114,13 +114,16 @@ func GetOutdatedPackages(installed map[string]desc.Desc, availablePackages cran.
...
-		if _, found := installed[pkgName]; found {
-			outdatedPackages = append(outdatedPackages, gpsr.OutdatedPackage {
-				Package:    pkgName,
-				OldVersion: installed[pkgName].Version,
-				NewVersion: pkgDl.Package.Version,
-			})
+		availableVersion := pkgDl.Package.Version
+
+		if installedPkg, found := installed[pkgName]; found {
+			if availableVersion != installedPkg.Version {
```
dpastoor: this still should use the desc version comparison, however this has been fixed upstream already in add-update feature so will not fix it now since that branch should merge quickly after

Issue #65, opened 2019-04-03
pkgr does not respect non-default R versions


dpastoor:
even when setting RPath in config, the initialization of a new RSettings which gets passed to the installers is not consuming the configuration

PR #66, opened 2019-04-03
Fix/rpath

dpastoor:
in response to #65 this fixes setting non-default R paths as for the execution of R. 

Primarily focused on environments with multiple versions of R present


dpastoor:
@Dreznel review to acknowledge the change and reason.

We'll want to have this in part of the 0.3.0 test suite. Will probably accomplish with a docker container with multiple versions of R installed and to run that test in said container.

PR #70, opened 2019-04-14
change: updating documentation and altering the way the Logging:insta…

dpastoor:
…ll option works

Setting Logging:install to a value will no longer prevent pkgr from logging to the "all" log file when pkgr install is run.

### DIFF
--- a/README.md
+++ b/README.md
@@ -129,7 +129,9 @@ Cache: "path/to/global/cache"
 
 # can log the actions and outcomes to a file for debugging and auditing
 Logging:
-  File: pkgr-install.log
+  all: pkgr-log.log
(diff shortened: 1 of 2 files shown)


dpastoor:
@Dreznel to confirm, the behaviors that are now implemented/expected are, all will log all pkgr activities, however if install is defined it will _also_ log the install commands to a specific file?

eg if i do 

```
pkgr plan
pkgr install
```

this would cause both the plan and install logs to go to all, but just the install logs to go to install?

Dreznel:
Devin,

Correct. "all" log contains everything, regardless of what other options are set. "install" log contains only install commands if set.

Issue #77, opened 2019-04-18
update should also update dependencies


dpastoor:
issue with devtools install when using --update, showing that an out-of-date dependency causes an install failure

```
ERRO[0022] cmd output                                    exitCode=1 package=devtools stderr="* installing *source* package ‘devtools’ ...\n** package ‘devtools’ successfully unpacked and MD5 sums checked\n** R\n** inst\n** tests\n** byte-compile and prepare package for lazy loading\nError in loadNamespace(i, c(lib.loc, .libPaths()), versionCheck = vI[[i]]) : \n  namespace ‘fs’ 1.1.0 is being loaded, but >= 1.2.0 is required\nERROR: lazy loading failed for package ‘devtools’\n* removing ‘/private/var/folders/g_/0948z6350756j3yly26_zqnr0000gn/T/devtools’\n" stdout=
```

when tracing back upwards can see

```
INFO[0016] package already installed                     package=fs
```

and all the way up to the outdated package found message, only outdated top level packages present.

Dreznel:
Can you provide a sample environment (in a zip or something) that recreates this issue you saw?

Based on what I know, **all** packages that are already installed are checked for outdatedness when you run install --update. Following this, here's my train of thought as to why this should already work:

1. The desired behavior of pkgr is to only install the latest versions of any package. 
    * We have not yet built in functionality for installing old packages, and we have deemed that out of scope for this task.
2. It follows from (1) that we expect packages to only depend on the latest versions of other packages.
3. If pkgr scans every installed package for outdatedness prior to installing, then it shouldn't be possible for dependencies to remain out of date in a normal, CRAN-ish space, unless 1/2 are not true.
4. An example:
    * If package A is dependent on package DEP, and package A is installed, then package DEP is installed. 
    * If A is outdated, then the newest version of A should depend on the most recent version of DEP. 
    * When we scan all packages for outdatedness, package A will be and primed for reinstallation. Package DEP will be flagged and primed as well iff the latest version is not already installed.
    * Package installation should proceed as normal from there.

Dreznel:
commit ddd3550 makes the functionality as described above.

dpastoor:
tested in #74, where crayon is an outdated dep only (not user package) and is shown to be flagged for update

Issue #75, opened 2019-04-18
clearer information about what is there/would be done


dpastoor:
right now, the INFO level is too verbose

```
 ~/rpkgs  pkgr plan                                                                                                                                                                          3928ms  Wed Apr 17 21:35:55 2019
INFO[0000] Installation would launch 11 workers
INFO[0000] R Version 3.5.3
INFO[0000] found installed package                       package=proj4 source repo=CRAN version=1.0-8
INFO[0000] found installed package                       package=crosstalk source repo=CRAN version=1.0.0
INFO[0000] found installed package                       package=survminer source repo=CRAN version=0.4.3
INFO[0000] found installed package                       package=maps source repo=CRAN version=3.3.0
INFO[0000] found installed package                       package=reshape source repo=CRAN version=0.8.8
INFO[0000] found installed package                       package=BiocManager source repo=CRAN version=1.30.4
INFO[0000] found installed package                       package=stringi source repo=CRAN version=1.3.1
INFO[0000] found installed package                       package=joineRML source repo=CRAN version=0.4.2
INFO[0000] found installed package                       package=maptools source repo=CRAN version=0.9-5
INFO[0000] found installed package                       package=rmarkdown source repo=CRAN version=1.11
INFO[0000] found installed package                       package=IRdisplay source repo=CRAN version=0.7.0
INFO[0000] found installed package                       package=V8 source repo=CRAN version=2.0
INFO[0000] found installed package                       package=generics source repo=CRAN version=0.0.2
INFO[0000] found installed package                       package=gameofthrones source repo=CRAN version=1.0.0
INFO[0000] found installed package                       package=prodlim source repo=CRAN version=2018.04.18
...
```

however it also has very handy information for sure such as:

```
INFO[0000] Default package type:  binary
INFO[0000] 5:5 (binary:source) packages available in for gh_external from https://metrumresearchgroup.github.io/rpkgs/gh_external
INFO[0000] 3:3 (binary:source) packages available in for gh_dev from https://metrumresearchgroup.github.io/rpkgs/gh_dev
INFO[0000] 13656:13884 (binary:source) packages available in for CRAN from https://cran.microsoft.com/snapshot/2019-03-15
INFO[0000] package repository set                        pkg=RcppRoll repo=CRAN type=binary version=0.3.0
INFO[0000] package repository set                        pkg=changepoint repo=CRAN type=binary version=2.2.2
[...]

Issue #74, opened 2019-04-18
pkgr plan supports use of update flag/setting


dpastoor:
similar to #69 any install activity should also be supported via plan to see what would happen without needing to actually perform the install command.

# validation requirements

## update argument overrides config
Implemented in: #78 
#### automated tests:
N/A
#### manual tests:
* `pkgr plan` on integration_tests/outdated-packages-no-update should not update already installed packages

![image](https://user-images.githubusercontent.com/3196313/56843137-1e42e280-686b-11e9-8017-2ad7c8f6bbdf.png)

* `pkgr plan --update` on integration_tests/outdated-packages-no-update should update outdated packages 

![image](https://user-images.githubusercontent.com/3196313/56843163-4c282700-686b-11e9-9a4f-163c76b5a209.png)

## running `pkgr plan --update` (or the equivalent using the config) will list the packages that will be updated, as well as what version they are currently on and what they will update to.

Implemented in: #78 
#### automated tests:
N/A
#### manual tests:
* `pkgr plan --update` on integration_tests/outdated_packages

![image](https://user-images.githubusercontent.com/3196313/56842986-7aa50280-6869-11e9-8b2d-d58719d3f12e.png)

## running `pkgr plan` without the update setting will warn you about outdated packages

Implemented in: #78 
#### automated tests:
N/A
#### manual tests:
* `pkgr plan` on integration_tests/outdated-packages-no-update should show outdated packages as a warning
* `pkgr plan --update` on integration_tests/outdated-packages-no-update should show outdated packages to be upgraded as info

![image](https://user-images.githubusercontent.com/3196313/56843057-36663200-686a-11e9-9701-734e96d50969.png)

Dreznel:
Plan  _does_ presently list outdated packages as warnings regardless of what options you pass in, but I will change this to make it more explicit.

dpastoor:
after correction for not trying to open non-folder elements in the package library to fail DESCRIPTION, this is complete

dpastoor:
fixed WARNING about missing DESCRIPTION from non-directory file in the package library (in the testing case a .gitignore).

functionality is good to go!

Issue #69, opened 2019-04-10
pkgr install --update option


Dreznel:
Create an update option for `pkgr install` that allows users to update outdated packages listed in their pkgr.yml file during the installation process.

* By default, pkgr should not update outdated packages
* With this option set, pkgr should attempt to update the outdated packages, but leave the current installation if the update-installation fails.
* For now, we assume that the user either updates everything or updates nothing -- we do not need to allow them to update specific packages.

# Validation Criteria

Implemented in #78, tested in #74 

##  If update operation fails, pkgr reinstates the previously-installed version.

Implemented in #78

dpastoor:
closing as so closely aligned to #74 that can use that for validation records

PR #78, opened 2019-04-18
add: update command for plan.

Dreznel:


Dreznel:
addresses #77 and #74

Dreznel:
I'm not strongly opinionated on this, but I personally prefer the word update. I would argue in favor of keeping it as update so that it matches thematically with update.packages.

As for number 2, I'm get on that.

And thanks! I agree, I tried really hard to be onboard with Go's naming conventions, but at least for pkgr, I'm finding it much easier to navigate after giving up those conventions.

Dreznel:
@dpastoor merge is done, there weren't any real conflicts.

Dreznel:
To validate:

#74
- [ ]  update flag can be set in config or passed in as an argument (--update)
- [ ] update argument overrides config
- [ ] running `pkgr plan --update` (or the equivalent using the config) will list the packages that will be updated, as well as what version they are currently on and what they will update to.
- [ ] running `pkgr plan` without the update setting will warn you about outdated packages

#61 
- [ ] for `pkgr install`, package name and version number are logged at an info level for prior-installed packages. 
- [ ] For `pkgr plan`, this information is logged at a debug level (to avoid verbosity)

#69 
- [ ] `pkgr install --update` (or the equivalent using the config file) installs all outdated packages specified to the latest available version in the specified repo.
- [x] Install with update does not currently check all available repos for most recent version -- it will use the first repo it finds. This is unideal but not in scope for this issue.
- [ ] If update operation fails, pkgr reinstates the previously-installed version.
- [x] If update operation is interrupted via a user command of some kind, we do not expect pkgr to be able to clean itself up.
- [ ] If nothing is specified in arguments or config, pkgr does not update.

#62 
- [ ] `pkgr install --loglevel debug` lists the size of package downloads in Megabytes.

#75 
- [ ] Package repositories to use are logged at an Info level for user-packages and a Debug level for dependencies.
- [ ] `pkgr plan/install` logs a brief "installation status" report detailing how many packages are required, how many are installed already, how many are outdated, and how many are to be installed or updated.

dpastoor:
attempting #74 looks really close, however noticed an artifact:

get a negative number of the number of packages to install, when no new packages need to be installed

image

Dreznel:
Started looking into the negative package numbers thing -- I think the issue is specifically when one removes packages from their config and then re-runs pkgr plan after already having installed.

I'll look into a fix for this.

dpastoor:
@Dreznel I think I know the root cause - namely if there are packages that have been installed to that folder from other entities than pkgr then the number of total packages can exceed that which would be installed/managed by pkgr. 

Hence, its not really a bug, but actually an unexpected feature.

dpastoor:
LGTM!

PR #79, opened 2019-04-19
change: refactored InstallPlan to carry its own download information.

Dreznel:
This could be done a little cleaner, but I wanted to throw this basic idea out there.

This was a potential refactoring I thought of when looking into #75 . My instincts tell me doing this will be helpful for differentiating between dependencies and user-level packages in the logs, but I haven't quite connected the dots yet.

### DIFF
--- a/gpsr/dependencies.go
+++ b/gpsr/dependencies.go
@@ -3,6 +3,7 @@ package gpsr
 import (
 	"errors"
 	"fmt"
+	"github.com/metrumresearchgroup/pkgr/cran"
 
 	"github.com/deckarep/golang-set"
 )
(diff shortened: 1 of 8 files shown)


gpsr/dependencies.go
```diff
@@ -116,3 +117,19 @@ func (ip InstallPlan) InvertDependencies() map[string][]string {
 	}
 	return idb
 }
+
+func (ip *InstallPlan) Pack(pkgNexus *cran.PkgNexus) {
```
Dreznel: If I go through with this, I'll be sure to pick a better name than "Pack." The idea is that I'm packing the installPlan with its own PkgDl objects so that the install command doesn't have to recalculate these.

PR #93, opened 2019-05-09
fix/r-version-constraints

Dreznel:


dpastoor:
handled in #94 instead

PR #94, opened 2019-05-09
Fix/r version constraints

Dreznel:


dpastoor:
pkgr builds and makefile tests complete

PR #99, opened 2019-05-14
feat: remove old binary from temp directory to cleanup

dpastoor:


dpastoor:
the randomness is just because we need some entropy to create a random folder name. There isn't really a need for 12 vs 16 vs otherwise. With regards to using a timestamp, especially seconds, actually is not so unique in a parallel setting, especially on kick-off will try to install many packages at the same time.

Issue #97, opened 2019-05-13
support detection of all platforms available via CRAN


dpastoor:
each platform uses a different naming convention for the names of the created binaries from R CMD BUILD, hence we need to support detection of the platform to derive the associated binary name during install.

For example, on ubuntu it is ~ x86_64_pc-linux.tar.gz, whereas on redhat it is instead like x86_64_redhat-linux.tar.gz

This can be taken from R.version$platform from a call from R directly. This should be done, for performance reasons, in the same call as finding the R version number itself as well.

PR #101, opened 2019-05-15
Feature/platform detection

david-lyder:
support detection of all platforms

#97

### DIFF
--- a/rcmd/Rsettings.go
+++ b/rcmd/Rsettings.go
@@ -45,33 +45,65 @@ func GetRVersion(rs *RSettings) cran.RVersion {
 			log.Fatal("error getting R version info")
 			return cran.RVersion{}
 		}
-		lines := rp.ScanLines(res)
-		for _, line := range lines {
-			s := strings.Fields(line)
-			switch s[0] {
-			case "version.string":
-				rsp := strings.Split(strings.Replace(s[3], "\"", "", -1), ".")
-				if len(rsp) == 3 {
-					maj, _ := strconv.Atoi(rsp[0])
-					min, _ := strconv.Atoi(rsp[1])
-					pat, _ := strconv.Atoi(rsp[2])
-					// this should now make it so in the future it will be set so should only need to shell out to R once
-					rs.Version = cran.RVersion{
-						Major: maj,
-						Minor: min,
-						Patch: pat,
-					}
-				} else {
-					log.Fatal("error getting R version")
-				}
-			case "platform":
-				// set platform in Rsettings for future access
-				rs.Platform = s[1]
-			default:
-			}
-		}
+
+		rs.Version, rs.Platform = parseVersionData(res)
+
+		// lines := rp.ScanLines(res)
+		// for _, line := range lines {
+		// 	s := strings.Fields(line)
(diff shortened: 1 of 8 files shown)


rcmd/Rsettings.go
```diff
@@ -33,32 +33,42 @@ func (rs RSettings) R() string {
...
 // to set that value, while also returning the RVersion.
 // This will keep any program using rs from needing to shell out multiple times
 func GetRVersion(rs *RSettings) cran.RVersion {
 	if rs.Version.ToString() == "0.0" {
-		res, err := RunR(afero.NewOsFs(), "", *rs, "paste0(R.Version()$major,'.',R.Version()$minor)", "")
+		res, err := RunR(afero.NewOsFs(), "", *rs, "version", "")
 		if err != nil {
-			log.Fatal("error getting R version")
+			log.Fatal("error getting R version info")
 			return cran.RVersion{}
```
dpastoor: don't need a return here, since will have fatally exited already

rcmd/Rsettings_test.go
```diff
@@ -36,3 +40,53 @@ func TestLibPathsEnv(t *testing.T) {
 		}
 	}
 }
+
+func TestGetRVersion(t *testing.T) {
+	var rVersionTests = []struct {
+		version    cran.RVersion
+		platform   string
```
dpastoor: First, we could refactor out the parsing of information to a separate function that can be tested in isolation of the actual runtime

something like
```
version, platform := parseRMetadata([]string)
```

then instead of having each one of these as individual elements of a struct, you could do a slice of strings or a single string with line breaks to more accurately represent how data gets fed in

rcmd/helpers.go
```diff
@@ -5,32 +5,43 @@ import (
...
 	case "linux":
-		return fmt.Sprintf("%s_%s_R_x86_64-pc-linux-gnu.tar.gz", pkg, version)
+		return fmt.Sprintf("%s_%s_R_%s.tar.gz", pkg, version, platform)
 	case "windows":
 		return fmt.Sprintf("%s_%s.zip", pkg, version)
 	default:
-		fmt.Println("platform not supported for binary detection")
-		return ""
+		log.Fatal("platform not supported for binary detection")
+		return ("")
```
dpastoor: no need for a return here as same with fatal 

have a empty return after the switch for the compiler happiness

rcmd/helpers_test.go
```diff
@@ -0,0 +1,62 @@
...
+
+func TestBinaryExt(t *testing.T) {
+	tests := []struct {
+		os       string
+		path     string
+		expected string
+	}{
+		{
+			os:       "linux",
+			path:     "/var/tmp/gz",
```
dpastoor: background: 

starting path will be something like: `path/to/<package>_<version>.tar.gz`

but when R CMD BUILD --binary is run, it will spit out a binary that is named specific to the platform/OS.

So we need to convert the tarball path into what we'd want to look for binary-wise.

```
tar.gz --> .tgz for mac
tar.gz --> .zip for windows
```

dpastoor:
LGTM

Issue #96, opened 2019-05-13
add/rm packages directly from the command line


dpastoor:
# validation requirements

## can add a package from cli

Implemented in: #102 

#### automated tests:
N/A
## user is alerted if package already exists when trying to add

Implemented in: #102 

#### automated tests:

N/A

#### manual tests:

* ran `add` with pkg `R6` on integration_tests/simple, expecting info message

david-lyder:
Test cases:
1. View usage/help for both add and remove.  Expect long and short description.
2. Add a single package, without --install. Expect pkgr.yml file to have new package in Packages section.
3. Remove a single package. Expect package to not be in Packages section of pkgr.yml file.
4. Add multiple packages, without --install. Expect pkgr.yml file to have new packages in Packages section.
5. Remove a multiple packages. Expect packages to not be in Packages section of pkgr.yml file.
6. Add a single package, with --install. Expect pkgr.yml file to have new package in Packages section and package folder to exist in test-library folder.
7. Add a multiple packages, with --install. Expect pkgr.yml file to have new packages in Packages section and package folders to exist in test-library folder.

Note: "remove" does not remove package folders from test-library. If folders exist, they must be manually removed to test "--install"

david-lyder:
Test 1. add and remove usage:
<img width="1027" alt="Screen Shot 2019-05-24 at 11 31 52 AM" src="https://user-images.githubusercontent.com/47984902/58339469-aa6b0a00-7e17-11e9-8ee0-24f005383558.png">

<img width="1027" alt="Screen Shot 2019-05-24 at 11 34 11 AM" src="https://user-images.githubusercontent.com/47984902/58339578-e605d400-7e17-11e9-957e-39104d5d3ef0.png">

david-lyder:
Test 2: Add a single package, without --install. Expect pkgr.yml file to have new package in Packages section.
<img width="1025" alt="Screen Shot 2019-05-24 at 12 06 16 PM" src="https://user-images.githubusercontent.com/47984902/58341525-6d554680-7e1c-11e9-8208-596e48982f5f.png">

david-lyder:
Test 3: Remove a single package. Expect package to not be in Packages section of pkgr.yml file.
<img width="1025" alt="Screen Shot 2019-05-24 at 12 07 54 PM" src="https://user-images.githubusercontent.com/47984902/58341613-a4c3f300-7e1c-11e9-9791-cb8f5fa6637c.png">

[...]

PR #102, opened 2019-05-17
Feature/add remove

david-lyder:
implements #96 

#96


cmd/add.go
```diff
@@ -0,0 +1,70 @@
...
+
+	if len(args) == 0 {
+		ccmd.Help()
+		os.Exit(0)
+	}
+
+	err := configlib.AddPackage(args[0])
+	if err != nil {
+		log.Fatalf("%s", err)
+	} else if install {
```
dpastoor: no need for an else - just have a second if statement

cmd/add.go
```diff
@@ -0,0 +1,70 @@
...
+		rInstall(nil, nil)
```
dpastoor: calling rInstall before calling initConfig?

cmd/add.go
```diff
@@ -0,0 +1,70 @@
...
+	// Initialize log and start time.
+	initAddLog()
+	startTime := time.Now()
+
+	if len(args) == 0 {
+		ccmd.Help()
+		os.Exit(0)
+	}
+
+	err := configlib.AddPackage(args[0])
```
dpastoor: why not range over all packages so the api naturally can be extended to `pkgr add pkgX pkgY etc`

configlib/config.go
```diff
@@ -66,3 +69,92 @@ func IsCustomizationSet(key string, elems []interface{}, elem string) bool {
...
+	i := 0
+	lines := bytes.Split(yf, []byte("\n"))
+	for _, line := range lines {
+		i++
+		out = append(out, line...)
+		if i < len(lines) {
+			out = append(out, []byte("\n")...)
+		}
+		if bytes.HasPrefix(line, []byte("Packages:")) {
+			out = append(out, []byte("  - "+packageName)...)
```
dpastoor: why not add the newline after pkgname in a single []bytes, instead of twice?

configlib/config.go
```diff
@@ -66,3 +69,92 @@ func IsCustomizationSet(key string, elems []interface{}, elem string) bool {
...
+	if err != nil {
+		return err
+	}
+
+	var out []byte
+	i := 0
+	lines := bytes.Split(yf, []byte("\n"))
+	for _, line := range lines {
+		i++
+		if bytes.HasPrefix(line, []byte("  - "+packageName)) {
```
dpastoor: this might not work with different whitespacing, should probably trim whitespace before checking for prefix

```
Packages:
- pkgX
```

and

```
Packages:
  - pkgX
```

are both valid

cmd/add.go
david-lyder: yup, nice catch on the whitespace, thanks

cmd/add.go
david-lyder: yikes! a bug.

cmd/add.go
david-lyder: Sure we can do that, let me know and I will implement and update the usage statement. Same for "remove"?

configlib/config.go
david-lyder: It's all about not writing the newline for the last line. The NL is written if it's not the last line of the file, or if we are adding the package.
One important unit test is to add a package, then remove the same package, and then have the same exact file (without ignoring whitespace)

configlib/config.go
david-lyder: agreed, change added with comment

cmd/add.go
dpastoor: doesn't initConfig still need to be called before anything else? eg initAddLog is expecting cfg.Logging.All/Overwrite to be set, doesn't that happen in initConfig?

cmd/add.go
david-lyder: added this feature

dpastoor:
testing this, one concern I had was whether or not whitespace differences between how add works and other pkgs could cause problems. This was sadly the case

In the case you have pkgs with no spacing, and add a new pkg (testaddpkg), the resulting yml is created:

```
Packages:
  - testaddpkg
- janitor
- BiocManager
- bookdown
- stringr
```

this however causes a panic

```
panic: unable to parse config file with error (While parsing config: yaml: line 4: did not find expected key)
```

we either need to detect whitespace and match what the user is already using _or_ do some santitization so the process of running add/remove will automatically clean to specific way.

cmd/add_test.go
```diff
@@ -0,0 +1,132 @@
...
+	command := exec.Command(pkgr, args...)
+	command.Dir = test.ymlfolder
+	command.Stderr = &stderr
+
+	err := command.Run()
+	errstr := string(stderr.Bytes())
+	return errstr, err
+}
+
+func isDiff(filename string) (bool, string) {
```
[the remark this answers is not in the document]
david-lyder: I have an existing test that saves the file as a byte slice before and after & compares - that can be used here as well. Here, I was looking to do an actual file compare, so I thought why not git? Especially since that was the key criteria I was looking for after the test: no recognized file changes.  I'll change the diff to a byte-compare.

rcmd/Rsettings.go
```diff
@@ -21,16 +22,26 @@ func NewRSettings(rPath string) RSettings {
 }
 
 // R provides a cleaned path to the R executable
-func (rs RSettings) R() string {
-	if rs.Rpath == "" {
-		return ("R")
+func (rs RSettings) R(os string) string {
+	r := "R"
+	if os == "windows" {
```
dpastoor: isn't this also fixed in 41, no need to double set it

david-lyder: It's needed if rs.Rpath is not empty and does not have the extension

PR #106, opened 2019-05-22
format yaml

dpastoor:
This PR forces yaml to be formatted prior to adding a package. This will autoformat to the correct spacing of 

```
Packages:
  - pkgX
  - pkgY
```

so the add command will properly work.

I needed to add yaml v3 which also allows us to allow to retain comments as this is not in the v2 that is pulled in via cobra at the moment


dpastoor:
@david-lyder please note changes and merge this into the add_remove branch before we finish that feature

dpastoor:
namely, this may break your existing tests as it will be forceably reformatting yaml files now so might need to rethink how those tests are performed

cmd/add.go
```diff
@@ -60,7 +60,6 @@ func rAdd(ccmd *cobra.Command, args []string) error {
 	}
 
 	if install {
-		initConfig()
```
dpastoor: while talking to David, this actually needs to stay, as need to re-initialize config to pull the written changes back in.

david-lyder:
looks good, will put initConfig back and create test for the yml formating

PR #119, opened 2019-05-31
Fix/r version detection

dpastoor:


dpastoor:
@Dreznel you might want to take a look at this. I needed to refactor the code to clean it up and eliminate some subtle bugs introduced in the clean command.

david-lyder:
I am OK with the first 3 changes. 
.Net offers a handy constant called Environment.Newline, of course the go community bashes the idea saying the user should know the file type. Perhaps, but I still like the concept of Envrionment.Newline because it will work in most cases, and highlights the cases where it does not work.

dpastoor:
Thats really not the (complete) issue in this case, as we wouldn't want to do Environment.Newline type thing because the point is it could be _either_ \r\n or \n, this way we can handle both gracefully.

FYI, if we're actually going line by line and not just splitting, bufio.Scanner will normalize that

There are some interesting convo's here https://github.com/golang/go/issues/28822

particularly mattn, who does some really good windows stuff in his support overall (he supports like the go-sqlite package and some other high profile stuff)

cmd/plan.go
```diff
@@ -157,7 +157,11 @@ func planInstall(rv cran.RVersion) (*cran.PkgNexus, gpsr.InstallPlan) {
 		for _, mp := range availableUserPackages.Missing {
 			log.Warnln("did you mean one of: ", model.Suggestions(mp, false))
 		}
-		os.Exit(1)
+		if exitOnMissing {
+			os.Exit(1)
```
Dreznel: Would log.Fatal be more appropriate?

cmd/plan.go
```diff
@@ -68,7 +68,7 @@ func plan(cmd *cobra.Command, args []string) error {
 	return nil
 }
 
-func planInstall(rv cran.RVersion) (*cran.PkgNexus, gpsr.InstallPlan) {
+func planInstall(rv cran.RVersion, exitOnMissing bool) (*cran.PkgNexus, gpsr.InstallPlan) {
```
Dreznel: should exitOnMissing be an argument here or a setting in the config file? Just some food for thought.

dpastoor: no, its an internal knob that needs to be tuned for specific ways we want to use planInstall. Namely, this is really an artifact of the fact that plan install is probably too monolithic. We need some way of getting back a pkgNexus even if downstream solves fail, so this is the quick fix. 

We don't want the user mucking with this (eg config) because it could cause very un-intended behavior. Eg to install we _must_ have a full plan, so not existing would cause incorrect behavior (a blank install plan that might exit with a "success")

Issue #116, opened 2019-05-30
new lockfile configuration option


dpastoor:
initial proposal is to have a top level configuration option:

```
Lockfile:
	Type: packrat | renv | pkgr
```

When the type is packrat AND library is not set, then install to `packrat/lib/<platform>/<rversion>`

dpastoor:
initially, can no-op renv/pkgr

david-lyder:
**Requirements:**
Support "packrat"-like folder structure (packrat/lib/<platform>/<rversion>) when pkgr.yml file:
1. Does not define "Library"
and
2. Defines Lockfile Type as "packrat"

david-lyder:
**Implementation proposal**
1. Define new type in PkgrConfig to support Lockfile, etc.
2. It's easy and tempting to overwrite cfg.Library if we are in "packrat mode', but it does not feel legit to overwrite viper marshall.  

cfg.Library is called in 3 locations, we can add logic in each location to use cfg.Library or a packrat path, or we can build a function\property to return the appropriate value.  

I lean toward a function in cmd\utils.go, or maybe in structs.go. It may not be a bad pattern to have "property-like" getters and setters in structs.go

dpastoor:
I am wondering if we need such "dynamism" in having a LibraryPath method that would have logic each time it is invoked. It would also require version and platform and/or the RSettings struct, based on your examples, to be carried around externally any place it was called. 

My gut is to say, given that this is a pretty static derivation, why not just do this logic in the constructor and be "done" with it, however that is at odds with how we are currently unmarshalling the cfg struct directly. That said, the more I think about it, per our convo on the phone, the more I'm thinking we may be doing some pre-processing after unmarshalling, but before the struct does anything else (eg variable replacement/injection), so creating an official constructor and doing it there could be a nice stepping stone to that.

dpastoor:
https://github.com/metrumresearchgroup/pkgr/blob/master/cmd/root.go#L108

I could imagine this as instead being something like

```
NewConfig(&cfg)
```

where NewConfig did the unmarshalling, then also started marching through and doing additional refinements, for example, checking if lockfile is of type packrat, and adjusting library, etc.

david-lyder:
I'm good with this idea ... the NewConfig function gives us a central location to make changes after the marshal

david-lyder:
For "pkgr plan", we log fatal if the Library path does not exist. Is it the user's responsibility to create Library path?

[...]

PR #120, opened 2019-05-31
feat: support packrat folder structure for installed packages, use Lo…

david-lyder:
…ckfile attribute in pkgr.yml

reference: #116

### DIFF
--- a/configlib/config.go
+++ b/configlib/config.go
@@ -14,9 +14,8 @@ import (
 	"github.com/spf13/viper"
 )
 
-// NewConfig create a new PkgrConfig
+// NewConfig initialize a PkgrConfig passed in by caller
 func NewConfig(cfg *PkgrConfig) {
-
 	_ = viper.Unmarshal(cfg)
 
 	if len(cfg.Library) == 0 {
--- a/configlib/config.go
+++ b/configlib/config.go
@@ -7,6 +7,7 @@ import (
 	"os"
 	"path/filepath"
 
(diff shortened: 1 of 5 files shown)


configlib/config.go
```diff
@@ -7,12 +7,48 @@ import (
...
+	return
+}
+
+func getLibraryPath(lockfileType, rpath, rversion, platform, library string) string {
+	switch lockfileType {
+	case "packrat":
+		library = filepath.Join("packrat", "lib", packratPlatform(platform), rversion)
+	case "renv":
+		rversion = fmt.Sprintf("R-%s", rversion)
+		library = filepath.Join("renv", "library", rversion, packratPlatform(platform))
```
dpastoor: with rEnv the r version is actually different. It just takes `R-3.5` whereas packrat takes 3.5.0 

we should probably switch getLibraryPath to take the rVersion itself then stringify each individually now.

```diff
-library = filepath.Join("packrat", "lib", packratPlatform(platform), rversion)
+library = filepath.Join("packrat", "lib", packratPlatform(platform), rv.ToFullString())
```

for packrat it would need rv.ToFullString() whereas renv needs rv.ToString()

david-lyder: how awesome that ToString() exists!

dpastoor: haha yeah there were enough places across the r ecosystem that it refers to one or the other so I added both of those from the get-go :-)

PR #122, opened 2019-06-03
Feature/detect package

david-lyder:


rcmd/structs.go
```diff
@@ -5,6 +5,9 @@ import (
 	"github.com/spf13/afero"
 )
 
+// VERSION is the current pkgr version
+var VERSION = "0.5.0-alpha"
```
dpastoor: The version should still be the CLI version. I was more thinking along the lines of

```
type ExecSettings struct {
	WorkDir string `json:"work_dir,omitempty"`
	PkgrVersion string `json:"pkgr_version,omitempty"`
}
```
then in the CLI you could do something along the lines of

```
// takes a dir + version
es := NewExecSettings(wd, VERSION)
```

cmd/inspect.go
```diff
@@ -94,11 +101,35 @@ func printDeps(deps map[string][]string, tree bool, ip gpsr.InstallPlan) {
 	}
 }
 
+func installedFromPackages() {
+	var pkgr, packrat, unknown []string
+	ip := GetInstalledPackages()
+	for k, v := range ip {
+		if len(v.PkgrVersion) == 0 {
+			packrat = append(packrat, k)
```
dpastoor: why packrat if PkgrVersion not present? 
Any reason not to do v.PkgrVersion == "", why check length?

cmd/plan.go
```diff
@@ -68,6 +68,11 @@ func plan(cmd *cobra.Command, args []string) error {
 	return nil
 }
 
+// GetInstalledPackages returns a list of PriorInstalledPackages
+func GetInstalledPackages() map[string]desc.Desc {
```
dpastoor: why have this, rather than just calling pacman.GetPriorInstalledPackages directly?

cmd/inspect.go
david-lyder: Interpreted from this statement:  "with the definition of not installed by pkgr being the metadata not detected in description file"

no reason for len, I think it did not compile the first time when using ==, probably due to an unrelated syntax error

cmd/plan.go
david-lyder: I felt it was cleaner to not pull pacman into Install.go.  Install.go feels like a top-level function that should call well-defined public functions - suitable for consumption without much context or inner knowledge of the package.  that _was_ my opinion, I now see that Install.go imports gpsr, so it's probably better to import pacman and call directly

rcmd/install.go
```diff
@@ -498,9 +513,51 @@ func InstallPackagePlan(
...
+		ir.Metadata.Metadata.Config.Repo.Name,
+		true)
+}
+
+func updateDescriptionInfo(filename, version, installType, repoURL, repo string, writeFile bool) ([]byte, error) {
+	var update []byte
+	fs := afero.NewOsFs()
+	dcf, err := afero.ReadFile(fs, filename)
+	if err == nil {
+		lines := bytes.Split(dcf, []byte("\n"))
```
dpastoor: checkout bufio.Scanner to break on lines but also handle windows vs unix lines

rcmd/install.go
```diff
@@ -498,9 +513,51 @@ func InstallPackagePlan(
...
+		ir.Metadata.Metadata.Config.Repo.URL,
+		ir.Metadata.Metadata.Config.Repo.Name,
+		true)
+}
+
+func updateDescriptionInfo(filename, version, installType, repoURL, repo string, writeFile bool) ([]byte, error) {
+	var update []byte
+	fs := afero.NewOsFs()
+	dcf, err := afero.ReadFile(fs, filename)
+	if err == nil {
```
dpastoor: this error is never handled, other than returning it, but in writeDescription you're not catching the error and doing anything about it

rcmd/install.go
```diff
@@ -498,9 +514,58 @@ func InstallPackagePlan(
...
 	if anyFailed {
 		log.Errorf("installation failed for packages: %s", strings.Join(failedPkgs, ", "))
 		return fmt.Errorf("failed installation for packages: %s", strings.Join(failedPkgs, ", "))
 	}
 	return nil
 }
+
+func writeDescriptionInfo(ir InstallRequest) {
+	_, err := updateDescriptionInfo(
+		filepath.Join(ir.InstallArgs.Library, ir.Package, "DESCRIPTION"),
```
dpastoor: I believe this library actually gets updated during the install, when installing through src. We need the final Library where it was ultimately installed. Packages installed from src go through src --> install in tmpdir to binary --> binary copied to binary cache --> installed through binary cache.


as of now all packages installed through src error with:

```
WARN[0038] DESCRIPTION file error:open /var/folders/g_/0948z6350756j3yly26_zqnr0000gn/T/GMWJGSAPGATL/devtools/DESCRIPTION: no such file or directory
```

dpastoor:
great - confirm that the change worked. Steps taken:

* installed R6, then manually deleted relevant lines from DESCRIPTION (confirmed they were in fact there) 
* saw pkgr warns on new install

```diff
devinp@hvdev:~/test-pkgr$ pkgr install
INFO[0000] R Version 3.6.0
INFO[0000] OS Platform x86_64-pc-linux-gnu
INFO[0000] found installed packages                      count=1
INFO[0000] Default package type:  source
INFO[0000] 5:5 (binary:source) packages available in for gh_external from https://metrumresearchgroup.github.io/rpkgs/gh_external
INFO[0000] 4:4 (binary:source) packages available in for gh_dev from https://metrumresearchgroup.github.io/rpkgs/gh_dev
INFO[0000] 14200:14200 (binary:source) packages available in for CRAN from https://cran.microsoft.com/snapshot/2019-05-12
INFO[0000] package repository set                        pkg=R6 relationship="user package" repo=CRAN type=source version=2.4.0
INFO[0000] package repository set                        pkg=glue relationship="user package" repo=CRAN type=source version=1.3.1
INFO[0000] package installation status                   installed=1 outdated=0 total_packages_required=2
INFO[0000] package installation targets                  to_install=1 to_update=0
INFO[0000] resolution time 296.947643ms
INFO[0000] downloading required packages within directory   dir=/home/devinp/test-pkgr/pkgcache
INFO[0000] downloading package                           package=glue
INFO[0000] all packages downloaded                       duration=535.996051ms
+WARN[0000] Packages not installed by pkgr:[R6]
INFO[0000] starting initial install
INFO[0003] Successfully Installed                        package=glue repo=CRAN version=1.3.1
INFO[0003] total install time                            duration=2.465169299s
INFO[0003] duration:3.317088448s
```

Then removed and reinstalled R6 and glue and saw no warnings.

PR #123, opened 2019-06-07
fix: prepend rPath to PATH in the environment used to call R

david-lyder:


rcmd/configure.go
```diff
@@ -53,7 +53,12 @@ func configureEnv(sysEnvVars []string, rs RSettings, pkg string) []string {
 				log.WithField("path", evs[1]).Debug("overriding system R_LIBS_USER")
 				continue
 			}
-
+			if evs[0] == "PATH" {
+				if rs.Rpath != "" && !strings.HasPrefix(evs[1], rs.Rpath) {
```
dpastoor: I think it actually should be the dirname of the path, not the exact path. Eg we don't need

`path/to/dir/R`, which is what Rpath is (path to r executable), instead we want that folder to be in the path.

This _may_ need to be experimented with, testing wise, to see if it actually fixes that issue.

Basically in the process is there a difference between doing

"exec R" where the R binary is in the path, vs "exec path/to/R". It _seemed_ that by not having that in the path, the error mentioned in the issue arose, as R itself could also not find some other key things. 

Hope that makes sense

david-lyder: yeah it makes sense. even though I noticed on my test that rPath was just "R", I still interpreted "rPath" as a path to the executable, not "the executable including its full path"

PR #147, opened 2019-08-14
wip: Feature/full rollback

Dreznel:
Proposed changes:

* Modify the installPlan to hold a list of pre-installed packages.
* Reset Library to original state using information from the installPlan

Overall process:
`pkgr install --update`:
* Uses the plan code to generate an installPlan
  * InstallPlan contains information about outdated packages and where their backups should be stored.
* Backup any packages to be updated into a folder named  \_\_OLD\_\_<package_name> (rename package directory).
* Attempt to install packages
* If the installation returns any errors, restore the Library to it's previous state by:
  * Deleting any packages that were not "pre-installed"
  * Restoring the backup-folders for the updated packages by removing the __OLD__ tag (using afero.Fs.Rename)


Dreznel:
I'm going to start refactoring from here to make sure this makes sense, and I will add automated tests as I go. If you'd like a head start, please take a look at the guide for integration_tests/rollback and tell me if there are cases you wish to try and add.

cmd/install.go
```diff
@@ -57,6 +58,8 @@ func rInstall(cmd *cobra.Command, args []string) error {
 	//  as well as a master install plan to guide our process.
 	_, installPlan := planInstall(rVersion, true)
 
+	//Make a backup of the current library
```
Dreznel: Will be removed, we went with a different strategy.

Dreznel:
Question I will look into: 

"preinstalled packages" in InstallPlan will end up containing the "outdated  packages" as well. This isn't affecting the overall process as far as I can tell, but that probably means we're calculating preinstalled packages a second time downstream in the process sometime _after_ we've moved the "to update" packages to the \_\_OLD\_\_ folders. What bothers me about this is that we're essentially defining two different sets of packages as "preinstalled": one that includes "to update" packages, and one that doesn't.

It's not a functional problem, but it might make things a bit confusing later on.

Dreznel:
As for the question above, the install process checks each package individually if it's installed prior to doing anything with it. In other words, the downstream process doesn't ever really generate a "list" of what's preinstalled, it just checks as it goes. While this is still kind of redundant, I don't think it's something we need to address at this time.

cmd/plan_test.go
```diff
@@ -88,3 +89,45 @@ func (suite *PlanTestSuite) TestGetPriorInstalledPackages_NoPreinstalledPackages
...
+	if err != nil {
+		return err
+	}
+
+	openedDir, err := fs.Open(src)
+	if err != nil {
+		return err
+	}
+	defer func() {
+		openedDir.Close() //If we
```
dpastoor: rather than deferring, especially since this is a recursively called function, this is an example where we should likely explicitly close as soon as we're done, which looks like after readdir in like 108

cmd/plan_test.go
```diff
@@ -88,3 +89,45 @@ func (suite *PlanTestSuite) TestGetPriorInstalledPackages_NoPreinstalledPackages
 func installedPackagesAreEqual(expected, actual desc.Desc) bool {
 	return expected.Package == actual.Package && expected.Version == actual.Version && expected.Repository == actual.Repository
 }
+
+func CopyDir(fs afero.Fs, src string, dst string) error {
```
dpastoor: should compare this against implementations such as https://github.com/otiai10/copy/blob/master/copy.go

Dreznel:
Note: found a bug related to outdated packages. In this branch (at least), if pkgs not related to the pkgr.yml file are already installed (i.e. they're not tracked by pkgr), and these packages have "updated" versions available in one of the listed repos, pkgr will move these packages into "OLD" folders without actually upgrading them.

I'll make sure that gets remedied in the next commit.

cmd/install.go
```diff
@@ -114,6 +121,40 @@ func rInstall(cmd *cobra.Command, args []string) error {
...
+
+	//figure out which packages we were even dealing with. Maybe this should be a unified list somewhere?
+	//as of August 14, 2019, the InstallPlan treats installed packages as if they were going to be installed again,
+	//and we simply figure out downstream that the installed packages shouldn't be run.
+	//Because of this, toInstall will end up containing info about all packages pkgr is managing, not just the ones
+	//to be installed.
+	toInstall := installPlan.StartingPackages
+	for depsList := range installPlan.DepDb {
+		toInstall = append(toInstall, depsList)
+	}
```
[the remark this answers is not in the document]
Dreznel: ! Good idea! Totally wasn't thinking in terms of methods.

cmd/plan.go
```diff
@@ -225,7 +224,7 @@ func planInstall(rv cran.RVersion, exitOnMissing bool) (*cran.PkgNexus, gpsr.Ins
 
 	log.WithFields(log.Fields{
 		"to_install":   toInstall,
-		"could_update": pkgsToUpdateCount,
```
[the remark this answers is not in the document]
Dreznel: I don't think this value was being assigned the way you're describing. Without the "--update" flag, it was reading 0. I'll double check this, but that was the rationale for changing it.

cmd/plan_test.go
Dreznel: I took a look. Our implementation doesn't handle symlinks, but overall they seem to be doing the same basic thing.

Since this function was recently demoted to a test-helper, and it has proven sufficient for the test cases that use it (most of which would always fail if this function didn't work), I think it's safe to leave it.

cmd/plan.go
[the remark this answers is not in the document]
Dreznel: Confirming what I said previously, when update is set to "false", the field formerly known as "could_update" was showing 0:

```
pkgr plan --update
INFO[0000] Installation would launch 11 workers         
INFO[0000] R Version 3.6.0                              
INFO[0000] found installed packages                      count=4
WARN[0000] Packages not installed by pkgr                packages="[R6 pillar Matrix crayon]"
INFO[0000] Default package installation type:  binary   
INFO[0000] 14273:14763 (binary:source) packages available in for CRAN from https://cran.microsoft.com/snapshot/2019-08-14 
INFO[0000] package will be updated                       installed_version=1.2.1 pkg=crayon update_version=1.3.4
INFO[0000] package will be updated                       installed_version=2.0 pkg=R6 update_version=2.4.0
INFO[0000] package will be updated                       installed_version=1.2.1 pkg=pillar update_version=1.4.2
INFO[0000] package will be updated                       installed_version=1.2-11 pkg=Matrix update_version=1.2-17
INFO[0000] package installation status                   installed=4 not_from_pkgr=4 outdated=4 total_packages_required=1
INFO[0000] package installation sources                  CRAN=1
INFO[0000] package installation plan                     to_install=1 to_update=4
INFO[0000] resolution time 245.511611ms

                 
johncarlos: ~/go/src/github.com/metrumresearchgroup/pkgr/integration_tests/outdated-pkgs$: pkgr plan --update=false
INFO[0000] Installation would launch 11 workers         
INFO[0000] R Version 3.6.0                              
INFO[0000] found installed packages                      count=4
WARN[0000] Packages not installed by pkgr                packages="[crayon R6 pillar Matrix]"
INFO[0000] Default package installation type:  binary   
INFO[0000] 14273:14763 (binary:source) packages available in for CRAN from https://cran.microsoft.com/snapshot/2019-08-14 
WARN[0000] outdated package found                        installed_version=1.2.1 pkg=crayon update_version=1.3.4
WARN[0000] outdated package found                        installed_version=2.0 pkg=R6 update_version=2.4.0
WARN[0000] outdated package found                        installed_version=1.2.1 pkg=pillar update_version=1.4.2
WARN[0000] outdated package found                        installed_version=1.2-11 pkg=Matrix update_version=1.2-17
INFO[0000] package installation status                   installed=4 not_from_pkgr=4 outdated=4 total_packages_required=1
INFO[0000] package installation sources                  CRAN=1
INFO[0000] package installation plan                     to_install=1 to_update=0
INFO[0000] resolution time 239.157699ms   
```
If it's easy to change it to the behavior you described, I'll fix it an change it back.

Dreznel:
(Next on my list is to add unit tests where I can)

pacman/utils.go
```diff
@@ -165,6 +165,30 @@ func tagOldInstallation(fileSystem afero.Fs, libraryPath string, outdatedPackage
...
 
+func CleanUpdateBackups(fileSystem afero.Fs, packageBackupInfo []UpdateAttempt) error {
+	if len(packageBackupInfo) == 0 {
+		log.Debug("Not update-packages to restore.")
+		return nil
+	}
+
+	for _, info := range packageBackupInfo {
+
+		_, err1 := fileSystem.Stat(info.BackupPackageDirectory) // Checking existence
```
dpastoor: can just check existence with https://godoc.org/github.com/spf13/afero#Afero.Exists

rollback/types.go
```diff
@@ -0,0 +1,64 @@
...
+
+type RollbackPlan struct {
+	AllPackages []string
+	NewPackages []string
+	UpdateRollbacks []pacman.UpdateAttempt
+	InstallPlan gpsr.InstallPlan
+	Library string
+}
+
+//Constructor
```
dpastoor: these should have comments as they are exported - for example

```
// CreateRollbackPlan creates a rollback plan to....
```

dpastoor: btw - does the goland linter not complain for not commenting on exported functions/methods?

rollback/types.go
```diff
@@ -0,0 +1,64 @@
...
+			opFiltered = append(opFiltered, op)
+		}
+	}
+
+	// Wrap this function for now until we're ready to move it over.
+	updateAttempts := pacman.PreparePackagesForUpdate(fs, library, opFiltered)
+	rp.UpdateRollbacks = updateAttempts
+}
+
+func DiscernNewPackages(toInstallPackageNames []string, preinstalledPackages map[string]desc.Desc) []string {
```
dpastoor: does this need to be exported? seems like the public interface is CreateRollbackPlan?

rollback/types.go
```diff
@@ -0,0 +1,64 @@
...
+	newPackages := make([]string, 0)
```
dpastoor: for consistency generally initialize a new variable with empty length (the default) by just declaring it

var newPackages []string

rollback/types.go
```diff
@@ -0,0 +1,64 @@
...
+	NewPackages []string
+	UpdateRollbacks []pacman.UpdateAttempt
+	InstallPlan gpsr.InstallPlan
+	Library string
+}
+
+//Constructor
+func CreateRollbackPlan (library string, installPlan gpsr.InstallPlan, preinstalledPackages map[string]desc.Desc) RollbackPlan {
+	ap := installPlan.GetAllPackages()
+	np := DiscernNewPackages(installPlan.GetAllPackages(), preinstalledPackages)
```
dpastoor: why not pass ap in, instead of calling installPlan.GetAllPackages again?

rollback/types.go
Dreznel: Looks like it doesn't, maybe that's a setting I can tweak somewhere.

rollback/types.go
Dreznel: Sure thing. Funnily enough, I was actually doing it _this_ way for consistency -- we don't use var declarations much, and so I was aiming to avoid using them (I remember reading a recommendation somewhere that said aim to use var declarations XOR := operators, but I agree with your recommendation here)

PR #149, opened 2019-08-15
refactor/unit-test

david-lyder:
test: create function and unit test for SetCustomizations

### DIFF
--- a/configlib/config.go
+++ b/configlib/config.go
@@ -204,3 +204,13 @@ func remove(ymlfile string, packageName string) error {
 	}
 	return nil
 }
+
+// SetCustomizations ... set ENV values in Rsettings
+func SetCustomizations(cfg PkgrConfig, rSettings *rcmd.RSettings) {
+	pkgCustomizations := cfg.Customizations.Packages
+	for n, v := range pkgCustomizations {
+		if v.Env != nil {
+			rSettings.PkgEnvVars[n] = v.Env
+		}
+	}
+}
--- a/cmd/install.go
+++ b/cmd/install.go
@@ -20,6 +20,7 @@ import (
 
 	"github.com/spf13/viper"
 
+	"github.com/metrumresearchgroup/pkgr/configlib"
 	"github.com/metrumresearchgroup/pkgr/cran"
 	"github.com/metrumresearchgroup/pkgr/logger"
 	"github.com/metrumresearchgroup/pkgr/pacman"
@@ -34,7 +35,7 @@ var installCmd = &cobra.Command{
 	Short: "install a package",
 	Long: `
 	install a package
- `,
+`,
 	RunE: rInstall,
 }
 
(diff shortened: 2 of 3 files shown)


configlib/config.go
```diff
@@ -204,3 +204,13 @@ func remove(ymlfile string, packageName string) error {
 	}
 	return nil
 }
+
+// SetCustomizations ... set ENV values in Rsettings
+func SetCustomizations(cfg PkgrConfig, rSettings *rcmd.RSettings) {
```
dpastoor: if we're going to be just modifying the rSettings, why not have setCustomizations be a method on it?

```
rs.SetCustomizations(cfg)
```

dpastoor: alternatively, if we keep this a function, I still vastly prefer the functional style then, give an object, return a new object, and in that case I would also change the order of args

```
rs = SetCustomizations(rs, cfg)
```

david-lyder: OK, I'll look into creating a method

configlib/config.go
```diff
@@ -206,11 +206,12 @@ func remove(ymlfile string, packageName string) error {
 }
 
 // SetCustomizations ... set ENV values in Rsettings
-func SetCustomizations(cfg PkgrConfig, rSettings *rcmd.RSettings) {
+func SetCustomizations(cfg PkgrConfig, rSettings rcmd.RSettings) rcmd.RSettings {
```
dpastoor: generally, when returning a modified an object we'd want the object itself to be the first arg

SetCustomizations(rSettings, cfg)...

david-lyder: done.

I usually put "in"s first and then the "out"s; in this case rSettings is kinda an "out". (Some languages have an out keyword to make it clear)

david-lyder:
The mystery of why the unit test "configlib\viper_test.go" fails when using "go test ./...", but passes using all other "go test" invocations (by package or by test) has been solved.

"go test ./..." runs tests in parallel. The yml file used in the tests was modified in "cmd/add_test.go" just before it was read by "configlib\viper_test.go".

Running the test with the flag "-p 1", which sets the number of parallel processed to one, succeeds. (go test -p 1 ./...)

Note: the "-p" flag is passed to the binary during execution, it is not used by "go test". (Running "go test" first generates a new binary that incorporates the unit tests, and then executes the generated binary). By default, the number of parallel processes used matches the number of cores, so without setting "-p 1" the number of processes was greater than one.  There is another flag used by "go test" named -parallel. This flag _is_ used by "go test" and allows tests to run in parallel as well as take advantage of parallel functions within the code. Interestingly, running "go test -parallel 1" also succeeds. I suspect the "-1" is passed when the generated binary is executed.  (The value of "parallel" defaults to the env var GOMAXPROCS, which was not defined in this case.)

Obviously passing the flag "-p 1" is not a good solution. Not using any file-system files during testing would be nice, but I feel the code that interacts with Viper is best tested using an an actual file. One (popular) solution is to pass strings to the functions instead of files. However, formatting yml in a go string can be a challenge - especially with respect to white space. I think it's easier to confidently create the needed test-cases using files. I like the pattern @Dreznel uses elsewhere in the code that copies a "golden" file. We could copy the golden yml file to a random temp folder, manipulate the copied yml and delete it when done. This would prevent the tests from clashing over the yml file.

Issue #170, opened 2019-10-31
cap number of threads to max of 8 unless explicitly set in the yml


dpastoor:
Past 8 threads and R can run into suggested dependencies webs that cause locking or other problems. Below that number does not have a problem. Running pkgr with a naive config on larger machines causes subtle problems that can be avoided this way.

## Acceptance Criteria:

* When no `threads` option is set in pkgr.yml or passed via command line, pkgr will use a number of threads equal to the number of CPUs available, or 8, whichever is lower.
* When the `threads` option is set in `pkgr.yml` or passed in through the command line, the value given will *always* be respected (no capping at 8 in this case).

Dreznel:
## Acceptance Criteria:

* User can set a 'max_threads' global option in pkgr.yml to limit the number of threads pkgr uses to 8
* Without any option set, pkgr will automatically cap number of threads to 8.

## Not in scope
* Command line argument to set max threads

Dreznel:
There is logic in the code that reduces the worker-count by 1 when the number of threads is two or higher. Given this, if we are "capping" the user at 8 threads (unless they request otherwise), should we keep this logic and limit them to 7 workers?

dpastoor:
That logic should only kick in if the number of threads was not explicitly set and actually I think we can remove that logic anyway.

I put that there originally in case the install process pegged the CPU as to not completely saturate and lock up the system. Fact is there is enough non CPU stuff going on during install, so that isn't a concern.

Hence, Let's remove that decrement logic all together

PR #174, opened 2019-11-06
change: unless user specifies otherwise, number of threads is now cap…

Dreznel:
…ped at 8.

references #170

We'll need to change the merge branch for this request once there's a release/v1.0.0 branch.


dpastoor:
changes reviewed in #175 and can be rolled into that branch, so will close this out as all commits present there.

Issue #173, opened 2019-11-05
Library should be created if does not exist


dpastoor:
This is a major request that have gotten numerous times. The original behavior was to error of the library did not exist, for two reasons:

1) we didn't have to worry about the permissions of that lib folder, that was up to the user to make sure was correct.

2) they couldn't accidentally make a lib in a place they weren't expecting

This has proven to be much less of an issue, and instead more painful to have to go make it yourself first, lots of

mkdir <lib> & pkgr install... 

## proposed change

* new top level parameter (accessible via command line and yml file) called `Strict` with a default of false

If strict is set to true, then the Lib must be created, else it will create. This will also position use to be able to tune other more consequential changes in the future. 

## Testing

- install pkg to lib folder when exists or not exists with strict on/off
- if strict off or folder exists should install, if strict on and folder doesn't exist should error with a informative message

Dreznel:
Counting the info under "proposed change" as Acceptance Criteria.

Dreznel:
A couple of questions as I'm working on this:

* Should `pkgr plan` indicate that the library will be created? I'm assuming yes.
* If the Library is created but a package fails to install, should the library be deleted? Assuming yes.

@dpastoor please confirm. I'm trying to determine whether or not I should store a "libraryCreated" variable in installPlan.

dpastoor:
hmm good questions.

I think both of these are reasonable choices and would proceed forward as such.

Issue #171, opened 2019-10-31
in install message list number of remaining packages 


dpastoor:
As a progress bar, per #129 is not really feasible with the logging, this can at least give a sense of what is remaining

```
INFO[0242] Successfully Installed                        package=stringr repo=MPN version=1.4.0
```

to now

```
INFO[0242] Successfully Installed                        package=stringr repo=MPN version=1.4.0 remaining=8
```

Dreznel:
## Acceptance Criteria
* pkgr lists a "remaining packages" count in the logs after each package is successfully installed.
* This information should be logged at an Info level.

Issue #126, opened 2019-07-03
improve the error messages when incorrectly specifying pkgr.yml path


dpastoor:
```
pkgr plan --library=/data/johnsmith/lib --config=/data/johnsmith
INFO[0000] Installation would launch 7 workers
INFO[0000] R Version 3.5.1
INFO[0000] OS Platform x86_64-pc-linux-gnu
INFO[0000] found installed packages                      count=99
PANI[0000] error getting pkgdb  Package database must contain at least one RepoUrl
panic: (*logrus.Entry) (0x9da7e0,0xc0004b38f0)
goroutine 1 [running]:
github.com/sirupsen/logrus.Entry.log(0xc0000ba1e0, 0xc00052c390, 0x0, 0x0, 0x0, 0x0, 0x0, 0x0, 0x0, 0x0, ...)
        /Users/devinpastoor/go/pkg/mod/github.com/sirupsen/logrus@v1.4.2/entry.go:239 +0x2ce
github.com/sirupsen/logrus.(*Entry).Log(0xc0004b2000, 0x0, 0xc00013ac48, 0x1, 0x1)
        /Users/devinpastoor/go/pkg/mod/github.com/sirupsen/logrus@v1.4.2/entry.go:268 +0xed
github.com/sirupsen/logrus.(*Entry).Logln(0xc0004b2000, 0xc000000000, 0xc00013afb0, 0x2, 0x2)
        /Users/devinpastoor/go/pkg/mod/github.com/sirupsen/logrus@v1.4.2/entry.go:359 +0xd3
github.com/sirupsen/logrus.(*Logger).Logln(0xc0000ba1e0, 0x0, 0xc00013afb0, 0x2, 0x2)
        /Users/devinpastoor/go/pkg/mod/github.com/sirupsen/logrus@v1.4.2/logger.go:239 +0x7f
github.com/sirupsen/logrus.(*Logger).Panicln(...)
        /Users/devinpastoor/go/pkg/mod/github.com/sirupsen/logrus@v1.4.2/logger.go:280
github.com/sirupsen/logrus.Panicln(...)
        /Users/devinpastoor/go/pkg/mod/github.com/sirupsen/logrus@v1.4.2/exported.go:219
github.com/metrumresearchgroup/pkgr/cmd.planInstall(0x3, 0x5, 0x1, 0x1, 0x1, 0x0, 0x0, 0x0, 0x0, 0x0, ...)
[...]

PR #175, opened 2019-11-06
Small features and fixes for 0.5.1/1.0.0

Dreznel:
Addresses the following issues: 
#173 - Library should be created if it doesn't exist
#170 - Cap threads to 8 when not specified
#171 - List number of remaining packages in install messages
#126 - Improve error messages when pkgr.yml not specified properly

### DIFF
--- a/cmd/install.go
+++ b/cmd/install.go
@@ -18,6 +18,7 @@ import (
 	"github.com/metrumresearchgroup/pkgr/pacman"
 	"github.com/metrumresearchgroup/pkgr/rollback"
 	"path/filepath"
+	"runtime"
 	"time"
 
 	"github.com/spf13/viper"
@@ -84,7 +85,7 @@ func rInstall(cmd *cobra.Command, args []string) error {
 	// leave at least 1 thread open for coordination, given more than 2 threads available.
 	// if only 2 available, will let the OS hypervisor coordinate some else would drop the
 	// install time too much for the little bit of additional coordination going on.
-	nworkers := GetWorkerCount()
+	nworkers := getWorkerCount(viper.GetInt("threads"), runtime.NumCPU())
 
 	// Process any customizations set in the yaml config file for individual packages.
 	// Set ENV values in rSettings
--- a/cmd/install.go
+++ b/cmd/install.go
@@ -71,6 +71,7 @@ func rInstall(cmd *cobra.Command, args []string) error {
 		if err != nil {
 			log.WithFields(log.Fields{
 				"library" : cfg.Library,
+				"error" : err,
 			}).Fatal("could not create library directory")
 		}
 	}
--- a/cmd/install.go
+++ b/cmd/install.go
@@ -97,9 +97,7 @@ func rInstall(cmd *cobra.Command, args []string) error {
 	pkgInstallArgs.Library, _ = filepath.Abs(cfg.Library)
 
 	// Get the number of workers.
-	// leave at least 1 thread open for coordination, given more than 2 threads available.
-	// if only 2 available, will let the OS hypervisor coordinate some else would drop the
-	// install time too much for the little bit of additional coordination going on.
+	// Use number of user defined threads if set. Otherwise, use the number of CPUs (up to 8).
 	nworkers := getWorkerCount(viper.GetInt("threads"), runtime.NumCPU())
 
 	// Process any customizations set in the yaml config file for individual packages.
--- a/cmd/install.go
+++ b/cmd/install.go
@@ -60,6 +60,15 @@ func rInstall(cmd *cobra.Command, args []string) error {
 	//  as well as a master install plan to guide our process.
 	_, installPlan, rollbackPlan := planInstall(rVersion, true)
 
+	if installPlan.CreateLibrary {
+		err := fs.MkdirAll(cfg.Library, 0755)
+		if err != nil {
+			log.WithFields(log.Fields{
+				"library" : cfg.Library,
+			}).Fatal("could not create library directory")
+		}
+	}
+
 	if viper.GetBool("update") {
 		log.Info("update argument passed. staging packages for update...")
 		rollbackPlan.PreparePackagesForUpdate(fs, cfg.Library)
--- a/cmd/install.go
+++ b/cmd/install.go
@@ -61,6 +61,12 @@ func rInstall(cmd *cobra.Command, args []string) error {
 	_, installPlan, rollbackPlan := planInstall(rVersion, true)
 
 	if installPlan.CreateLibrary {
+		if cfg.Strict {
+			log.WithFields(log.Fields{
+				"library": cfg.Library,
+			}).Fatal("library directory must exist before running pkgr in strict mode -- halting execution")
+		}
+
 		err := fs.MkdirAll(cfg.Library, 0755)
 		if err != nil {
 			log.WithFields(log.Fields{
--- a/cmd/install.go
+++ b/cmd/install.go
@@ -84,7 +84,7 @@ func rInstall(cmd *cobra.Command, args []string) error {
 	// leave at least 1 thread open for coordination, given more than 2 threads available.
 	// if only 2 available, will let the OS hypervisor coordinate some else would drop the
 	// install time too much for the little bit of additional coordination going on.
-	nworkers := getWorkerCount()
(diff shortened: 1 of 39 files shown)


cmd/utils.go
```diff
@@ -28,21 +28,30 @@ func userCache(pc string) string {
 	return pkgrCacheDir
 }
 
-func getWorkerCount() int {
+func GetWorkerCount() int {
```
dpastoor: whats the point of having these as exported functions? eg GetWorkerCount vs getWorkerCount. Likewise, why just return another function and not use GetRestrictedWorkerCount directly. If you're going to abstract, you want some more logic to be at play.

cmd/utils_test.go
```diff
@@ -0,0 +1,56 @@
...
+	for _, test := range inputs {
+		executeTestGetRestrictedWorkerCount(t, test.context, test.threads, test.numCpus, test.expectedOutput)
+	}
+}
+
+func executeTestGetRestrictedWorkerCount(t *testing.T, context string, threads int, numCpus int, expectedOutput int) {
+	t.Log(fmt.Sprintf("context: %s", context))
+	actualResult := GetRestrictedWorkerCount(threads, numCpus)
+	assert.Equal(t, expectedOutput, actualResult)
+	t.Log("PASS")
```
dpastoor: no need to log PASS? any reason for that. If it doesn't fail, it by default passes

Dreznel: So I logged PASS mostly to make the output more easily readable while I was testing. Doing table-testing this way may be good for writing tests, but it makes their results a decent bit harder to parse for the developer in the IDE.

I'll remove it, but I think this warrants future consideration.

cmd/utils.go
Dreznel: The abstraction was to make unit-testing easier. I wanted to keep the "complexity" (however minimal) of calling runtime.numCPUs at this level of code rather than push it up a layer, as I think the upper layers are already a bit cluttered. I am still of this opinion, but do you want me to change it anyway?

As for exporting, that was me thinking this was Java and that I wouldn't be able to unit-test from another file unless it was exported. I'll change that, but I want to leave the abstraction.

cmd/plan.go
```diff
@@ -75,18 +76,39 @@ func plan(cmd *cobra.Command, args []string) error {
 func planInstall(rv cran.RVersion, exitOnMissing bool) (*cran.PkgNexus, gpsr.InstallPlan, rollback.RollbackPlan) {
 	startTime := time.Now()
 
-	installedPackages := pacman.GetPriorInstalledPackages(fs, cfg.Library)
-	installedPackageNames := extractNamesFromDesc(installedPackages)
-	log.WithField("count", len(installedPackages)).Info("found installed packages")
-	whereInstalledFrom := pacman.GetInstallers(installedPackages)
+	//Check library existence
+	libraryExists, _ := afero.DirExists(fs, cfg.Library)
```
dpastoor: always still worth logging the error and not swallowing it. This is a good example where panic works

configlib/structs.go
```diff
@@ -51,4 +51,5 @@ type PkgrConfig struct {
 	Logging        LogConfig           `yaml:"Logging,omitempty"`
 	Update         bool                `yaml:"Update,omitempty"`
 	Lockfile       Lockfile            `yaml:"Lockfile,omitempty"`
+	Strict				 bool								 `yaml:"Strict,omitempty"`
```
dpastoor: spacing

rcmd/install.go
```diff
@@ -5,6 +5,7 @@ import (
 	"bytes"
 	"errors"
 	"fmt"
+	"github.com/spf13/viper"
```
dpastoor: rcmd should not need to ever know about how configuration happens. All functions should be parameterized. Hence, instead should update the install api to take a update bool, that you give in the CLI/otherwise the viper.GetBool result

rcmd/install.go
```diff
@@ -380,17 +381,30 @@ func InstallPackagePlan(
...
 	startTime := time.Now()
+
 	// for now this will only be updated in the Update function
 	// however if it may be concurrently accessed should consider a syncmap implementation
 	installedPkgs := make(map[string]bool)
 	// if packages ID'd as ready to install signal so can push them only the queue
 	shouldInstall := make(chan string)
 	anyFailed := false
+
+	toInstall := plan.GetNumPackagesToInstall(viper.GetBool("update"))
```
dpastoor: see above comment

rcmd/install.go
```diff
@@ -406,11 +420,13 @@ func InstallPackagePlan(
 				log.WithFields(log.Fields{"binary": iu.BinaryPath, "src": pkg.Path}).Debug(iu.Package)
 
 				if iu.Result.ExitCode != -999 {
+					packagesNeeded = packagesNeeded - 1
 					log.WithFields(log.Fields{
 						"package": iu.Package,
-						"version": pkg.Metadata.Package.Version,
-						"repo":    pkg.Metadata.Config.Repo.Name,
-					}).Info("Successfully Installed")
+						"package version": pkg.Metadata.Package.Version,
```
dpastoor: no spaces in log fields - `package_version` etc. Though this is also something we want to maintain consistency on. In this case we know the package, why need to be so explicit and call it package version - can just say `version`. This also should be consistent with the plan() nomenclature

rollback/operations.go
```diff
@@ -13,6 +13,19 @@ func RollbackPackageEnvironment(fileSystem afero.Fs, rbp RollbackPlan) error {
 	//reset packages
 	logrus.Trace("Resetting package environment")
 
+	if rbp.InstallPlan.CreateLibrary {
+		err0 := fileSystem.RemoveAll(rbp.Library)
+
+		if err0 != nil {
+			logrus.WithFields(logrus.Fields{
+				"library": rbp.Library,
```
dpastoor: error can be a field `error = err0`

rcmd/install.go
Dreznel: So this was a strange one. Log.WithFields lists fields in alphabetical order, and I couldn't find another way to have "packages remaining" listed last. If you're okay with having "remaining" listed at the far left end of the fields, I'll change it back. Otherwise it's going to need to go on its own line or as part of the actual logging statement.

cmd/install.go
```diff
@@ -84,7 +85,7 @@ func rInstall(cmd *cobra.Command, args []string) error {
 	// leave at least 1 thread open for coordination, given more than 2 threads available.
 	// if only 2 available, will let the OS hypervisor coordinate some else would drop the
 	// install time too much for the little bit of additional coordination going on.
-	nworkers := GetWorkerCount()
+	nworkers := getWorkerCount(viper.GetInt("threads"), runtime.NumCPU())
```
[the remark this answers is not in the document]
Dreznel: Good catch, will change.

rcmd/install.go
dpastoor: far left of the fields is fine. If we need to eventually, we can write our own formatter, https://github.com/sirupsen/logrus/issues/329 For now, really we just need to make it in. It definitely doesn't need to be on its own line/actual logging statement

rcmd/install.go
Dreznel: Agreed. The only reason I didn't was because it felt odd to me to add an "update" parameter to this function when it's only used for logging.

I think what I'll try to do is store the "update" boolean in the InstallPlan, then this method can just read from there instead of taking it as an argument.

cmd/plan.go
```diff
@@ -77,7 +77,13 @@ func planInstall(rv cran.RVersion, exitOnMissing bool) (*cran.PkgNexus, gpsr.Ins
 	startTime := time.Now()
 
 	//Check library existence
-	libraryExists, _ := afero.DirExists(fs, cfg.Library)
+	libraryExists, err := afero.DirExists(fs, cfg.Library)
+
+	if err != nil {
+		log.WithFields(log.Fields{
```
dpastoor: panic(err)

dpastoor: after discussion, we should instead use log.Error to still communicate the error loudly but not crash as to stop the plan in case downstream elements still need to use plan info

dpastoor:
Lgtm!

PR #194, opened 2019-11-08
Test Preparation

Dreznel:
This branch focuses on getting our integration and unit tests ready to be formally run. 

Also contains a  fixes for issues  and #196  #197


Dreznel:
@dpastoor  please see d4fc4de and let me know what you think of this for our screenshot strategy. I think it's a little clunky, but it seems to be the direction we're heading.

dpastoor:
Nice work. I think you're definitely in the right direction. A couple thoughts: 

1) I don't know how I feel about checking the results in. In particular the impact this will have on the repository size over time. One alternative to this would be instead use the wiki as the place to store this. The wiki itself is a git repository and can be edited locally, though there is not really a concept of review and approval

2) The other we could do, to keep repo size down, would be to never merge the branch with the screenshots. This would cause a slow growth of NOMERGE branches though, which might make navigating more difficult over time.

3) If approvals for particular issues need different reviewers, it would get messy as to how to approve sub elements.

I am thinking that we keep this same structure, but rather than have the results.md as a file, instead it is just an issue. That way it can be reviewed, updated, etc. Github provides audit trail of edits, we can see who was reviewed (assigned) etc.

Dreznel:
Unit tests passing (except the one we discussed in Slack):

image

Dreznel:
mixed-source integration test successfully completes (this is a labor-intensive  integration test that works well as a sanity check because of its breadth):
image

Issue #164, opened 2019-10-09
updating description file should ONLY retain the repo installed from, not the original repo


dpastoor:
else this causes issues with snapshotting

For example in the original description given:

```
Repository: CRAN
```

after installing from MPN it will look like:

```
Repository: CRAN MPN
```

```
Repository: MPN
```

when detected that already Repository Field exists, it can change to `OriginRepository`

Dreznel:
## Acceptance Criteria:
* For any package P installed by pkgr, the DESCRIPTION file of P should have a "Repository" entry that matches exactly the user-defined name of the repository that pkgr retrieved the package from, with no additional text.

PR #196, opened 2019-11-22
Rewriting functions for updating DESCRIPTION files

Dreznel:
Addresses #164 

## Note
I branched off of tests/test-tags because I didn't want to lose the altered tests I've been writing in case I make more alterations during this process. So this pull request is currently pointing there to avoid confusion during review. I'll remove the validation folder before final merge.

## TODO
* Code review

## TODONE:
* Get function to work properly with newlines in a cross-platform way.
* Write unit tests

#### First commit message
The functions work, but newlines are not preserved so the new DESCRIPTION file is just one-line. Some quick-ish research seems to indicate that Go does not have a reliable cross-platform solution to newline characters (most of the functions that should do this apparently still use \n in the background, at least as of 2018). I hoped to get around this by using bufio WriteString line-for-line, but somewhere along the way newlines are being dropped. I'll keep plugging away at this.


rcmd/testsite/golden/simple/pkgr.yml
```diff
@@ -0,0 +1,12 @@
+Version: 1
```
Dreznel: This whole golden directory can be removed, they were from when I was trying to update the older tests.

rcmd/install.go
```diff
@@ -534,44 +534,123 @@ func writeDescriptionInfo(ir InstallRequest, ia InstallArgs) {
...
+	newLines = append(newLines, "PkgrVersion: "+version)
+	newLines = append(newLines, "PkgrInstallType: "+installType)
+	newLines = append(newLines, "PkgrRepositoryURL: "+repoURL)
+
+	return newLines
 }
+
+// function to inspect the current operating system and attempt to determine appropriate line-endings.
+// if Windows, returns "\r\n". Otherwise, returns "\n".
+func getEOLcharacter() string {
```
Dreznel: @dpastoor A few things here:
1) Newlines are trimmed by Scanner (including carriage returns, if they're present), so simply trying to modify the lines in-place doesn't work unless we're scanning in as bytes.
2) Additionally, we have to add a new line for OriginalRepository no matter how you look at it, so we need a way to access the correct newline character for the OS. fmt.Fprint() and such uses \n behind the scenes (I checked), so that wasn't an option. 
3) Since newlines are a problem elsewhere in the code, I don't expect this to be the final resting place for this function, but I'm not sure where it ought to go to avoid circular dependencies, and I didn't want to commit to making a new top-level folder. So basically, we should discuss this.

rcmd/install_test.go
```diff
@@ -39,49 +36,203 @@ func TestInstallArgs(t *testing.T) {
...
+			installType: "binary",
+			repoURL: "https://www.fakecranrepos.org",
+			repo: "AlCRAN_Mandragoran",
 		},
 	}
+	fmt.Println(tests["test1"])
+	for testName, test := range tests {
+		t.Run(testName, func(t *testing.T) {
+			results := updateDescriptionInfoByLines(test.startingLines, test.version, test.installType, test.repoURL, test.repo)
+			assert.Len(t, results, 7)
```
dpastoor: so this actually breaks the entire table driven test abstraction. The idea is that the assertions in the loop should be applicable for each test - this is a "table" test with just one test, thus these pass. But what we should have a number of types of tests that flex the different scenarios

rcmd/install_test.go
```diff
@@ -39,49 +36,203 @@ func TestInstallArgs(t *testing.T) {
...
+				"PkgrInstallType: binary", // matches final result
+				"PkgrRepositoryURL: https://www.fakecranrepos.org",
+			},
+			version: "pkgr0.0.test",
+			installType: "binary",
+			repoURL: "https://www.fakecranrepos.org",
+			repo: "CRAN",
+		},
+	}
+	fmt.Println(tests["test1"])
```
dpastoor: drop this?

rcmd/install_test.go
Dreznel: I will not fix this for now since it shouldn't impact validation docs (as per our Slack discussion), but I understand the rationale and philosophy here and will write tests that way going forward.

rcmd/install.go
Dreznel: Irrelevent -- we are comfortable always writing out using \n, and we are using ReadLinesFS from goutils now to more robustly parse _in_ the files line-by-line.

PR #199, opened 2019-12-03
Censor Environment Variables

Dreznel:


rcmd/configure.go
```diff
@@ -80,3 +86,16 @@ func configureEnv(sysEnvVars []string, rs RSettings, pkg string) []string {
 
 	return envVars
 }
+
+
+// Returns a constant set of env vars to be hidden in logs.
+// Calling function may pass in additional values to include in the censor list.
+func censoredEnvVars(add []string) []string {
+	censoredVars := []string{"GITHUB_TOKEN", "GITHUB_PAT", "GHE_TOKEN", "GHE_PAT", "AWS_ACCESS_KEY_ID", "AWS_SECRET_KEY",}
```
dpastoor: it would make more sense to have censoredEnvVars be a map - that way you get instant lookup and do not need to use go-funk to scan the array each time.

Dreznel:
Just acknowledging that this is on my radar.

rcmd/configure_test.go
```diff
@@ -31,61 +31,113 @@ func TestConfigureArgs(t *testing.T) {
...
 			[]string{"R_LIBS_USER=original/path", "MISC2=bar"},
-			[]string{"DPLYR_ENV=true", "R_LIBS_SITE=path/to/install/lib", "MISC2=bar"},
+			[]string{"DPLYR_ENV=true", "MISC2=bar", "R_LIBS_SITE=path/to/install/lib"},
 		},
 		{
 			"R_LIBS_SITE and R_LIBS_USER env set",
 			"",
 			[]string{"R_LIBS_USER=original/path", "R_LIBS_SITE=original/site/path", "MISC2=bar"},
-			[]string{"R_LIBS_SITE=path/to/install/lib", "MISC2=bar"},
+			[]string{"MISC2=bar", "R_LIBS_SITE=path/to/install/lib"},
```
Dreznel: @dpastoor After merging in the unit test fixes from the other branch, I'm taking another glance at these and realizing that I don't fully understand the "business logic" here, and that makes me worried that these tests are misconfigured. 

Specifically, it looks like `R_LIBS_USER` is dropped or replaced with `R_LIBS_SITE` in all cases. According to this comment:
```
			// we don't want to track the order of these anyway since they should take priority in the end
			// R_LIBS_USER takes precedence over R_LIBS_SITE
			// so will cause the loading characteristics to
			// not be representative of the hierarchy specified
			// in Library/Libpaths in the pkgr configuration
			// we only want R_LIBS_SITE set to control all relevant library paths for the user to
```

`R_LIBS_USER` is supposed to get precedence, which I would assume means we want it to appear later in the final environment variable list. Our tests do not currently reflect this behavior.

I'm going to proceed writing tests as if the current tests are accurate, but please let me know if I need to make updates.

Dreznel:
Here is a screenshot of the most recent unit-test output:

image

Please see my comment above, but if there are no concerns there then this should be good for final review and merge.

dpastoor:
well, as I cycle back to this - I have done a very poor job code reviewing this. This actually is not the intent, by replacing the values completely, though the logs themselves will show censored, the actual values themselves are gone too for underlying functionality that needs to access the true values.

This needs to be rewritten. This hasn't rear'd its head as pkgr should never really need any of those values to do what it needs as it exists, but also those values are no longer present

PR #206, opened 2019-12-12
NOMERGE: Release/v0.5.1 tests

Dreznel:

PR #207, opened 2019-12-12
change: fixing broken unit tests for add and delete.

Dreznel:
Since PR #206 contains testing artifacts, I'm creating this branch as a way to update automated tests without adding in all the testing artifacts.

That said, as I'm writing this, I'm realizing that we also need to check in updates to the guide.md files in the integration_tests, so maybe we'll just make a branch at the end of the effort that excludes the `validation` folder but merges in everything else.


dpastoor:
LGTM - can you post a snippet showing the tests are passing (on your machine) then it'll be good to merge

Dreznel:
@dpastoor one test that was previously skipped was inadvertently fixed by some modifications, so I've also enabled that one. All unit tests are passing now:

image

PR #210, opened 2019-12-18
wip: Release documentation

Dreznel:
Need to include:
* List of supported features
* Descriptions of any odd behavior (such as logging config)
* Descriptions of commands and flags.

### DIFF
--- a/MANUAL.md
+++ b/MANUAL.md
@@ -7,16 +7,6 @@ All `pkgr` commands follow the format `pkgr <command> <flags>`
 `pkgr` or `pkgr --help`
 * Prints a list of top level pkgr Commands
 
-#### Universal Flags (can be applied to any command)
-* `--config [path]`: Path to the `pkgr.yml` file, if not `./pkgr.yml`.
-* `--preview`: Enable preview mode [[ unclear on what this does ]]
-* `--debug`: Enable debug mode (developer only).
-* `--loglevel [info/debug/trace]`: Set the log level to "info", "debug", or "trace".
-* `--update` or `--update=false`: Set pkgr to update outdated packages it finds in the user's library.
-* `--threads [number]`: Set pkgr to use up to the specified number of threads.
-* `--strict` or `--strict=false`: Enable or disable strict mode (see notes)
-* `--help`: List information about a command.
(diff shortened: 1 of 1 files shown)


MANUAL.md
```diff
@@ -0,0 +1,202 @@
...
+  ```
+
+## Default Settings
+Below are the default settings for non-required pkgr configurations.
+  * Debug mode: disabled
+  * Preview mode: disabled
+  * Strict mode: disabled
+  * Loglevel: “info”
+  * Rpath: “R”
+  * Threads: Number of processors available on system
```
dpastoor: threads not processors - eg on my 6 core mac --> will default to 12 threads.

MANUAL.md
```diff
@@ -0,0 +1,202 @@
...
+  * Cache: System temp directories
```
dpastoor: User cache directory (not system) - under the hood this uses ~ os.UserCacheDir() which is I believe like ~/.cache/... on unix systems

Dreznel: > this is looking really good/helpful. The one thing is I think this needs to either be in the wiki _or_ the readme. I doubt people will find the manual (I'm not aware of other repos that have a manual.md), so if we're going to go off the readme, we might as well use a "real" system - aka the wiki, that can be maintained/extended as large as we need it to be.

I agree that the Wiki is the place to put this ultimately, I'll transfer the information over there once we've gotten everything straight with the testing docs, as those are currently cluttering it up.

Dreznel:
> this is looking really good/helpful. The one thing is I think this needs to either be in the wiki _or_ the readme. I doubt people will find the manual (I'm not aware of other repos that have a manual.md), so if we're going to go off the readme, we might as well use a "real" system - aka the wiki, that can be maintained/extended as large as we need it to be.

I agree that the Wiki is the place to put this ultimately, I'll transfer the information over there once we've gotten everything straight with the testing docs, as those are currently cluttering it up.

Dreznel:
I have made the requested changes and ported the MANUAL over to [the wiki](https://github.com/metrumresearchgroup/pkgr/wiki/user_manual).

Issue #217, opened 2020-01-13
Feature: As a user, I can disable the "rollback" feature and let pkgr install all packages that do not fail


Dreznel:
# Summary
Users must be able to "turn off" the rollback feature, which will allow pkgr to partially complete a `pkgr install` command when one or more packages are failing to install.

# Tests:
tbd

PR #218, opened 2020-01-13
Feature: Disable Rollback

Dreznel:
Addresses #217


Dreznel:
@dpastoor I need to add unit tests, but please take a look at the `rollback-disabled` guide.md file, particularly at the "Current behavior," and tell me if that's what you'd like.

Dreznel:
Tests passing:
image

Dreznel:
Helpful info and current behavior as validated by `TestInstallWithoutRollback`:

* crayon: **Normal, preinstalled** -- A "regular" package that is already installed and doesn't need updates.
* fansi: **Normal, not installed** -- A "regular" package that will install correctly.
* R6: **Outdated** -- An outdated package that successfully installs prior to error.
* Rcpp: **Outdated dependency of failure** -- An outdated package that the "fail" package relies on and will update.
* utf8: **Not installed by pkgr** -- A package that is not included at all in the pkgr file. We may want to remove this, but I figured we shouldn't be corrupting users' environments even if they're using pkgr in an unintended way.
* flaxtml: **Depends on fail package** -- A package that depends on xml2 (the package that fails to install). Dependencies for flaxtml should install, but flaxtml itself should fail to install because xml2 failed to install.
* RCurl: **Is a dependency of flaxtml** RCurl should still be installed despite the fact that xml2 -> flaxtml could not be installed.
* bitops: **Is a dependency of RCurl** Bitops should still be installed, because flaxtml should still be installed.

Dependency tree (via `pkgr install --deps`):
```
{
  "RCurl": [
    "bitops"
  ],
  "flatxml": [
    "Rcpp",
    "bitops",
    "RCurl",
    "xml2"
  ],
  "xml2": [
    "Rcpp"
  ]
}
```

## Current behavior observed:
`pkgr install`:
* R6 and Rcpp update successfully
* fansi, bitops, and RCurl are installed
* xml2 fails to install and says `ERRO[0000] installation failed for packages: xml2`     
* flaxtml does not install, but there is no message saying the installation failed.

Dreznel:
Changed base to develop, please merge.

PR #220, opened 2020-01-15
Automated integration test for disable-rollback feature

Dreznel:
I took a swing at writing an automated test for the rInstall function. It involved some refactoring, as well as the creation of a local repo.

The test takes a long time to run, so in practice we'd mark it as "Skip" in most cases and only bust it out when necessary (or when there's time). Definitely faster than re-running the manual integration test every time though, which means I can be flexible with requirements with much less overhead.

Also, I haven't written any helper-functions to make this into a pattern, but obviously this is something we could do. Once I have those, I can possibly rig up some kind of table-test to make this even neater.

In it's current form, consider this a "proof of concept"


cmd/install_test.go
```diff
@@ -0,0 +1,110 @@
...
+				"local58" : configlib.RepoConfig{
+					Type: "source",
+				},
+			},
+		},
+		//LibPaths: nil,
+		//Lockfile: nil,
+		Repos: []map[string]string{
+			{
+				"local58" : "/Users/johncarlos/go/src/github.com/metrumresearchgroup/pkgr/localrepos/bad-xml2",
```
Dreznel: This will need to be changed to something more generic.

PR #226, opened 2020-02-03
Feature/install tarball2

Dreznel:
A different, more simple approach to the tarball-install feature.

Requirements:
* Tarball must be able to be installed from a local repository or from a URL to that tarball.
* Tarball should be in the format of a built R package
* Dependencies of tarball should be installed during the initial install phase
* Tarball installation should be "highest priority", overriding previous installations of the same package.

Assumptions:
* Tarball paths will be specified in their own "Tarballs" list in pkgr.yml
* Tarballs will each have a DESCRIPTION file within them.
* Any packages to be installed via Tarball will NOT be dependencies of other packages specified anywhere in pkgr.yml.

Questions:
* If a package to be installed via tarball already exists/has already been installed, should pkgr always overwrite the existing installation?


cmd/plan.go
```diff
@@ -243,6 +259,134 @@ func planInstall(rv cran.RVersion, exitOnMissing bool) (*cran.PkgNexus, gpsr.Ins
 	return pkgNexus, installPlan, rollbackPlan
 }
 
+// Tarball manipulation code taken from https://gist.github.com/indraniel/1a91458984179ab4cf80 -- is there a built-in function that does this?
+func unpackTarballs(fs afero.Fs, cfg configlib.PkgrConfig) []desc.Dep {
```
dpastoor: rather than passing in the entire config - grab the things you need from the config as specific vars. Eg cache, Tarballs can be separate args. It makes it much more clear around what unPackTarballs needs, and makes it easier to isolate/test etc

gpsr/helpers.go
```diff
@@ -69,3 +69,11 @@ func appendToGraph(m Graph, d desc.Desc, dependencyConfigs InstallDeps, pkgNexus
 		}
 	}
 }
+
+
+// Left off here -- we can easily gather a list of Dep objects and try to append them,
+// but if we add the full-blown Desc objects, then we'll be putting the tarball package in the install plan.
+// We're probably just going to do a hacky workaround for this.
+func appendSingleDependencyToGraph(m Graph, d desc.Dep, pkgNexus *cran.PkgNexus) {
```
dpastoor: this function scaffold has two things that could be improved - first dependency is already singular - so could be appendDependencyToGraph(...). Furthermore, this could easily be generalized to take a variadic param of desc's, so could easily pass those one or many

```
appendDependenciesToGraph(m, pkgNexus, d)
appendDependenciesToGraph(m, pkgNexus, d1, d2, d3)
appendDependenciesToGraph(m, pkgNexus, listOfDeps...)
```

Dreznel:
FYI, I will need to clean this up significantly, but for now I'm just trying to get something working. The latest commit achieves the objective but spits out the binaryball somewhere random. I'm going to try and use InstallThroughBinary to address this.

Dreznel:
Note: merged from #230 as part of the latest batch, which is why there are so many commits.

Here is proof of testing for the latest:
image

Dreznel:
@dpastoor Ready for review, with the following caveats:
* I need to adjust the "packages remaining" counter to account for tarballs.
* I need to fix a bug where Tarballs installed via pkgr are flagged as "not installed by pkgr" on subsequent runs (unless you'd like to keep this? it might be good to call out that a tarball isn't coming from the "usual" pkgr workflow)
* I need to tighten up some of the log messages (let me know if there's anything that stands out to you)
* Some of the method arguments could be broken down (passing elements of cfg instead of cfg, etc.) but I may just leave these for now. The functions are high enough in the chain that we're not likely to unit-test them directly.

dpastoor:
> I need to fix a bug where Tarballs installed via pkgr are flagged as "not installed by pkgr" on subsequent runs (unless you'd like to keep this? it might be good to call out that a tarball isn't coming from the "usual" pkgr workflow)

We will still consider this installed by pkgr - the big concept that it should be is "we know who/how it was installed". Likewise installed by pkgr means we'll know some pkgr metadata is injected in the description file.

Dreznel:
@dpastoor ready for review

Tests passing:

image

Dreznel:
Tests pass after merging in tilde expansion:

image

Dreznel:
Those last two commits are actually significant.

In 43ab9e4, I realized that I was leveraging the cache for tarball installations without actually providing a way to clean it out via `pkgr clean cache`, so I addressed that. 

In b134e5c, I modified everything so that Tarball installations can be rolled back the same way anything else can.

Dreznel:
Question: should `pkgr load` attempt to load Tarballs?

cmd/load.go
```diff
@@ -0,0 +1,308 @@
...
+	if all {
+		toLoad = installPlan.GetAllPackages()
+	} else {
+		toLoad = userPackages
+	}
+
+	report := InitLoadReport(getRSessionMetadata(rs, rDir))
+
+	resultsChannel := make(chan LoadResult, len(toLoad))
+	sem := make(chan int, threads * 2)
```
dpastoor, line 91: think about something like....threads *2 given ncores on system is >=4 or threads*3 - i tried this by loading a load with concurrency of only 4 and i wasn't near saturating any CPU, so there is plenty of room at the top.

dpastoor: consider overriding with --load-threads=N

cmd/load.go
```diff
@@ -0,0 +1,308 @@
...
+	} else {
+		log.WithFields(log.Fields{
+			"working_directory": rDir,
+			"failures": report.Failures,
+		}).Error("some packages failed to load.")
+	}
+
+	if toJson {
+		//getRSessionMetadata(rs, rDir)
+		log.Info("printing load report as JSON")
```
dpastoor, line 141: one point of printing out json, is making. it parseable by json readers. By logging to info, you're going to corrupt the json of stdout

dpastoor: actually this will just be. ano-op because you set log level to fatal at top so can remove this message

cmd/load.go
```diff
@@ -0,0 +1,308 @@
...
+	return pkgLoadMetadata{
+		pkgPath,
+		pkgVersion,
+	}
+}
+
+// Helper function to run R commands and gather the outputs, using the set of arguments that `pkgr load` internal
+// operations require.
+func runRCmd(rExpression string, rs rcmd.RSettings, rDir string, reducedOutput bool) ([]string, []string, error) {
+	cmdArgs := []string{
```
dpastoor, line 283: lets add a `--no-restore` here

dpastoor: this is the only thing that _needs_ changing before we merge

cmd/cleanCache.go
```diff
@@ -93,6 +93,28 @@ func cleanCacheFolders() error {
 		return errors.New("'what? that's impossible! my logic is flawless!'")
 	}
 
+	// Clear the tarballs
+	log.Info("clearing unpacked tarballs from the cache")
+	for _, tgzFile := range cfg.Tarballs {
+		openedTgz, err := fs.Open(tgzFile)
```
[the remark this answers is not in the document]
dpastoor: and please close this

cmd/cleanCache.go
```diff
@@ -93,6 +93,28 @@ func cleanCacheFolders() error {
...
+		if err != nil {
+			log.WithField("file", tgzFile).Warn("could not find untarred directory in local cache for file")
+			continue
+		}
+		hashedDirectoryName, err := getHashedTarballName(openedTgz)
```
dpastoor, line 104: don't close before this

cmd/load.go
Dreznel: Created issue #251 to address in the future

Issue #228, opened 2020-02-07
pkgr load cli command


dpastoor:
loading all pkgs in a library as a sanity check on the fact that all are successfully installed (at least enough to load) can help catch some compilation and runtime bugs that don't present during installation.

In addition, switching R versions, workflow types, or other situations where the binaries have been on the system but are stale can be checked.

## expected behavior

to run all user defined packages in Packages

```
pkgr load
```

will run threads x 2 (since there are some definite IO bottlencecks) where in each subprocess will run the equivalent of

```
R -e "library(<pkg>)"
```

and capture the stdout, stderr, and error code and return normal logrus printed logs that contain the information specified in the json section below.

By default, all successful packages should log a success at the Debug level, Failed packages should present with an Error level, and at the end a summary statement at the info level providing the metadata + a summary

```
INFO[9999] "packages loaded"  checked=120 success = 120 failure=0
```

Hence, for a successful check, the log should be very quiet, however can see the individual results for each that would look like:

```
DEBU[9999] "dplyr"    version="0.8.4" path=... exitcode=0 stderr=... stdout=...
```

### all

```
pkgr load --all
```

will load _all_ packages in the dependency tree, not just the user defined packages

### json

```
pkgr load --json
```

At the _end_ it will return a **single** json object 

```json
{
  metadata: {
   rpath: path/to/R,
   r_version: "3.6.2",
   libpaths: [
      "path/to/lib",
      "other/path"
   ] 
  },
  results: [
    {
      package: "dplyr",
      version: "0.8.4",
      exit_code: 0,
      stderr: "",
      stdout: "",
      path: "path/to/install/folder"
    },
    {
      package: "tidyr",
      version: "1.0.0",
      exit_code: 0,
      stderr: "",
      stdout: "",
      path: "path/to/install/folder"
    },
  ]
}
```

note: there is no equivalent to getting all the runtime libpaths as of yet, this is not just the library per the install, but actually how the R session will see the libraries when its launched. Will need to parse a call to `.libPaths()` to get that.

## additional behaviors:

For install, pkgr by default starts in a highly isolated mode, when this launches, it should launch as the *user* would interact, so _not_ a --vanilla --slave --noprofile, etc session used when installing pkgs. 

## future considerations

should be able to provide a command line flag to change the wd from where the R session will start (to appropriately capture separate .Rprofile's or other things that are important to control where the R session launches, not where the pkgr.yml file lives).

Dreznel:
[...]

PR #230, opened 2020-02-17
Pkgr Load Command

Dreznel:
Pull request to address #228 

Branched off of the disable-rollback feature, #218, at commit d25082bae5ae008b6cc28622364cde637ba21b0f


dpastoor:
Before you go any further with this - I'd like to see a (small) writeup of what PoC'ing you've done to accomplish this - particularly outside of go, to demonstrate the expected inputs/outputs you're going to be handling.

In addition, you should connect that to what _already exists_ in the package.

There is a bit of wheel reinventing going o, for example:

https://github.com/metrumresearchgroup/pkgr/blob/master/rcmd/rp/lines.go#L12


In addition design of your functions - baking in the exact command to use, rather than providing a framework around being able to execute more R code and consume results. From a first glance, I'm not seeing the connection points around how to provide a reasonable simple, but flexible mechanism to collect the results we know we'll need - libPaths(), outputs of library calls, etc

dpastoor:
To be more explicit - lets start with just showing, in the terminal "by hand", what commands you're expecting to run, the expected outputs and ergonomics of each (for example exit code, stderr, stdout for packages with various degrees of failure/success). Then connect that back to the implementation approach.

Dreznel:
Thanks for the feedback, @dpastoor. I'll start working to incorporate it. I expected there to be some level of code-cleanup once I had everything in place, so I'll make sure to address these concerns.

I obviously missed the `ScanLines` function, but in other cases I _was_ looking for opportunities to reuse code, and I just didn't find many. I found the functions in `RunR.go`, but since they weren't being used in [the install function](https://github.com/metrumresearchgroup/pkgr/blob/128f4ae5de6cac839de138b96ba43c3fd0292d76/rcmd/install.go#L138), I decided to follow the pattern there and just generate individual commands for this purpose. This is just to say that if there were other places you saw me writing redundant code (aside from redundancies within `pkgr load` itself), please point them out. To mitigate risk, I try not to touch the `install` code unless I'm directly working on it, but now that I understand the exec package better, I'm comfortable refactoring out common functionality if you think that's appropriate here.

The commands I'm looking to run are pretty much:
* `R -q -e "library(<package>)"`
  * Can be silent or have console output. Error output goes to stderr.
  * Will have a 0 exit code if library was loaded successfully, will have a non-zero exit code otherwise.
  * We only need to check the exit code, everything else is just extra information to put in the JSON object. 
* `R -q -e "find.package('%s'); packageVersion('%s')"`
  * Could be split into two subfunctions, I just kept them together for brevity.
  * I use these to grab the installation path and the installed version of each package, both of which are returned in the stdout if there is no error.
  * If there _is_ an error, we can still finish the `load` operation -- this information is not crucial. We can simply say "problem fetching package path," and _that_ can tip off the user that something might be amiss for this particular package. Presumably, if the package was able to load successfully, then these two functions will also work.
  * We discussed possibly scanning the libPaths and parsing description files to accomplish this. I decided it would be better to do this from the R session explicitly so that we can be sure we're always returning what R itself finds. It's possible for the same package to be installed in two different libraries, in which case R will just use the one it finds first.
* `R -q -e ".libPaths()"
  * Probably won't fail unless R isn't installed. If it fails, then we can log the failure and move on, as this is just meta-information for the JSON report.
  * We use this to parse out the libPaths for logging/JSON purposes.

Dreznel:
@dpastoor Hey Devin, this is waiting on the next round of review. I could use help on the proper implementation of the parallelization

dpastoor:
so some initial comments:

1) your code is not reflecting your logic as you (correctly) where planning.

for example, you said above "We only need to check the exit code, everything else is just extra information to put in the JSON object." - that is correct, yet you have an enormous number of failing conditions because of the fact that packages will log startup messages as messages, which go to stderr not stdout.

```
"rsample": {
            "Package": "rsample",
            "Version": "‘0.0.5’",
            "Path": "\"/Users/devinp/rpkgs/3.6/rsample\"",
            "Exiterr": "",
            "Stdout": "\u003e library('rsample')\n\u003e\n\u003e",
            "Stderr": "Loading required package: tidyr",
            "Success": false
        },
```

2) output form is very sloppy. Version, Path, and Stdout all need cleaning up. I don't even know whats going on with the Path

```
"Version": "‘0.0.5’",
"Path": "\"/Users/devinp/rpkgs/3.6/rsample\"",
"Stdout": "\u003e library('rsample')\n\u003e\n\u003e",
```

3) the purpose of the json flag is to _literally_ only return the single json object. This also means suppressing log messages. The point is you should be able to pipe that to a json file or consume that in via a stream.

As you have it, its virtually useless printing out in the middle of a bunch of other logs.

image


A very simple rule of thumb to always follow:

1) make it work
2) make it correct
3) make it fast

I think you got so caught up in trying to implement the parallel piece (3) that you didn't first cover the first 2 responsibilities. Your ideas you laid out above are spot on, and virtually exactly the approach I would have taken as well. Lets first get those working correctly, then we'll come back to the parallel piece.

dpastoor:
Getting closer - one thing that is causing unnecessary difficulty is the use of `-q` vs `--slave`

```
~/rpkgs R --slave -e "2+2"                                                                                                                                                                 09:33:32
[1] 4
⋊> ~/rpkgs R -q -e "2+2"                                                                                                                                                                      09:33:45
> 2+2
[1] 4
> 
> 
```

this will keep the output minimized and not need to strip those extra `>` which is the encoding when you see things like `\u003e library('xtable')\n\u003e\n\u003e`

dpastoor:
In addition, this should make parsing the other metadata you need very easy:

```
~/rpkgs R --slave -e "as.character(packageVersion('tidyverse')); find.package('tidyverse')"                                                                                                09:36:38
[1] "1.3.0"
[1] "/Users/devinp/rpkgs/3.6/tidyverse"
```

Dreznel:
I managed to figure out how JSON encoders work, and so I think I'm going to keep that in since it's a bit more correct (and not too complicated), but I will also incorporate `slave`.

Dreznel:
I believe I have implemented the Semaphore pattern you described, but please take a look.

Also, an interesting thing I ran into, if you're in a repo with a big .RData file, we're trying to load that in with every R session we launch. We ought to pass in an argument to prevent this, as it makes pkgr load take way too long.

Dreznel:
Hey @dpastoor , we had a chat that somehow you had seen older code in your latest review. Could you give this another look and see what you think?

Dreznel:
Tests pass

image

Dreznel:
Suggesting we close this merge request, as it has been incorporated into #226

PR #240, opened 2020-03-07
Feature/recommended

dpastoor:


gpsr/helpers.go
```diff
@@ -3,10 +3,14 @@ package gpsr
...
+	log "github.com/sirupsen/logrus"
 )
 
-func isDefaultPackage(pkg string) bool {
-	_, exists := DefaultPackages[pkg]
+func isDefaultPackage(pkg string, noRecommended bool) bool {
+	pkgType, exists := DefaultPackages[pkg]
+	if exists && !noRecommended && pkgType == "recommended" {
+		return false
+	}
```
Dreznel: This function name is misleading now.

The previous behavior just told you whether or not a package was a default package. 

The new behavior sort of returns a decision on whether or not a package should be excluded as a default package.

I suggest renaming this function to `isExcluded`, `excludePackage`, `excludePackageAsDefault`,  or something to that effect.

dpastoor: I think that is reasonable

Issue #235, opened 2020-03-03
all things that could be paths should attempt to expand ~


dpastoor:
On initial parsing, the following should attempt to be expanded:

* library
* Repo paths, incase they are local repos rather than html

```
Library: ~/rpkgs
```

or 

```
Repos:
- local: "~/crandbs/somedb"
```

Dreznel:
@dpastoor 

This wasn't on my radar, but we still want this for the March release right?

dpastoor:
yes please.

Dreznel:
This is done and can be closed.

PR #243, opened 2020-03-17
Feature/expand tilde

Dreznel:
Branched off of #226 

Branch to handle issue #235 

Please pay no attention to the misspelled branch name.


Dreznel:
Integration test passes:

image

configlib/config.go
```diff
@@ -38,9 +38,61 @@ func NewConfig(cfg *PkgrConfig) {
...
+	}
+	return expanded
+}
+
+/// For a list of paths, expand the ~ at the beginning of each path to the home directory.
+/// consider any problems a fatal error.
+func expandTildes(paths []string) []string {
+	var expanded []string
+	for _, p := range paths {
+		newPath := expandTilde(p)
```
shairozan: tldr; Package not needed. Personal opinion, use go's user lookup tools

This package isn't required anymore. In around go 1.9, 'nix systems will use /etc/passwd and more to locate the user details:

[Golang user docs](https://golang.org/pkg/os/user/)

> For most Unix systems, this package has two internal implementations of resolving user and group ids to names. One is written in pure Go and parses /etc/passwd and /etc/group. The other is cgo-based and relies on the standard C library (libc) routines such as getpwuid_r and getgrnam_r.

While it's fine to use for simplicity (IE you've already used) it may make more sense to implement the user looking and collection of homedir in-application. Especially on a project with such a low code velocity as the go-homedir one.

configlib/config.go
```diff
@@ -38,9 +38,61 @@ func NewConfig(cfg *PkgrConfig) {
 		rVersion := rcmd.GetRVersion(&rs)
 		cfg.Library = getLibraryPath(cfg.Lockfile.Type, cfg.RPath, rVersion, rs.Platform, cfg.Library)
 	}
+
+	// For all cfg values that can be paths, make sure that ~ is expanded to the home directory.
```
shairozan: Nice implementation. Just basically inject the changes before they would be used.

configlib/config.go
```diff
@@ -38,9 +38,61 @@ func NewConfig(cfg *PkgrConfig) {
...
+		expanded = append(expanded, newPath)
+	}
+	return expanded
+}
+
+/// In the PkgrConfig object, Repos are stored as a list of key-value pairs.
+/// Keys are repo names and values are paths to those repos
+/// For each key-value pair, expand the prefix ~ to be the home directory, if applicable.
+/// consider any problems a fatal error.
+func expandTildesRepos(repos []map[string]string) []map[string]string {
```
shairozan: I don't see any unit tests for this? Can you provide one? Or point to where it is?

shairozan: Primarily because it's the most complicated logic change to the branch

Dreznel: Thanks for the review! 

I had intentionally omitted unit tests for this change because we're testing home directories via `~`, which aren't consistent across machines. I suspect this might be handleable via afero, but I haven't looked into it yet.

dpastoor: well one simple test can be, for example, to set a repo given ~/blah, then expect that the return string does not contain ~/, or as DJ mentioned above, use one of the other mechanisms to derive what homedir.Expand _should_ do. 

something like:

```
userDir := os.UserHomeDir() // this is made up
assert.equal(repos[toExpand], userDir + 'hard/coded/rest')
```

configlib/config.go
dpastoor: I see so basically now we can just use os.Current()

which returns a pointer to a:

```
type User struct {
    // Uid is the user ID.
    // On POSIX systems, this is a decimal number representing the uid.
    // On Windows, this is a security identifier (SID) in a string format.
    // On Plan 9, this is the contents of /dev/user.
    Uid string
    // Gid is the primary group ID.
    // On POSIX systems, this is a decimal number representing the gid.
    // On Windows, this is a SID in a string format.
    // On Plan 9, this is the contents of /dev/user.
    Gid string
    // Username is the login name.
    Username string
    // Name is the user's real or display name.
    // It might be blank.
    // On POSIX systems, this is the first (or only) entry in the GECOS field
    // list.
    // On Windows, this is the user's display name.
    // On Plan 9, this is the contents of /dev/user.
    Name string
    // HomeDir is the path to the user's home directory (if they have one).
    HomeDir string
}
```

and just ask for `user.HomeDir`

That said, FWIW, I'd argue the low velocity makes go-homedir an easy choice since its well tested, a fixture of use, and keeps us from having to maintain (an admittedly small surface area) additional code. If anything, I'd fork it, change the implementation if we felt the need, then started using our version, rather than dropping it in each application.

configlib/config.go
```diff
@@ -38,9 +38,61 @@ func NewConfig(cfg *PkgrConfig) {
...
+	var expanded []map[string]string
+	//expanded := make(map[string]string)
+	for _, keyValuePair := range repos {
+		kvpExpanded := make(map[string]string)
```
dpastoor: whats the reasoning for initializing a new map each time then growing a new object vs just modifying the values in the original map?

configlib/config.go
Dreznel: The other nice thing is that homedir automatically takes care of the actual string interpolation, which we'd have to do ourselves otherwise. Not particularly complicated, but it's something.

configlib/config.go
Dreznel: I'm trying to be consistent with how the Repos structure in PkgrConfig is implemented. IIUC, The Repos structure is a slice of maps, with each individual map being a single key-value pair, so we're replicating that exactly here. That's why this particular function is called `expandTildesRepos` instead of something like `expandTildesMap`

dpastoor: What?

I'm saying why create a _new_ map each time. Why not modify the original value in the existing map

repo[key] = homedir.Expand(repo[key])

Dreznel: Ah, my bad. 

I don't have an explicit reason for this, but I think instinctively I wanted to treat the Repos object in PkgrConfig as immutable to avoid any bugs that might've come up as a result of modifying by reference.

I'll make the change.

Dreznel: Walking back a bit,

Unless you're strongly opinionated on this, I'd actually rather leave it as it is so that all of the "expandTilde" functions work the same way.

```
	cfg.Library = expandTilde(cfg.Library)
	cfg.RPath = expandTilde(cfg.RPath)
	cfg.Tarballs = expandTildes(cfg.Tarballs)
	cfg.Repos = expandTildesRepos(cfg.Repos)
	cfg.Logging.All = expandTilde(cfg.Logging.All)
	cfg.Logging.Install = expandTilde(cfg.Logging.Install)
	cfg.Cache = expandTilde(cfg.Cache)
```
I find the above code nice and consistent, simple to reason about, etc, whereas having just one of those lines suddenly be different
```
	cfg.Library = expandTilde(cfg.Library)
	cfg.RPath = expandTilde(cfg.RPath)
	cfg.Tarballs = expandTildes(cfg.Tarballs)
	expandTildesRepos(cfg.Repos)
	cfg.Logging.All = expandTilde(cfg.Logging.All)
	cfg.Logging.Install = expandTilde(cfg.Logging.Install)
	cfg.Cache = expandTilde(cfg.Cache)
```
Seems kinda strange. I don't think the performance/memory implications in this case are big enough to worry about, either.

PR #248, opened 2020-03-31
Feature/tarball downloads

Dreznel:
I could clean this up a bit more, but I figured I'll submit it for commentary first. 

This feature adds the ability to specify remote tarballs for "AdditionalPackage"  installation.

### DIFF
--- a/cmd/clean.go
+++ b/cmd/clean.go
@@ -48,11 +48,12 @@ func clean(cmd *cobra.Command, args []string) error {
 	}
 	if cleanAll {
 		fmt.Println("Cleaning all.")
-		err = cleanCacheFolders()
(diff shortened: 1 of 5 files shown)


Dreznel:
Big things to note are:

- We use the cache as more of a working tmp directory than an actual cache.
- All Tarballs are downloaded into a special folder in the cache so that they can be easily cleared on a `pkgr clean` command.
- Tarballs are downloaded as part of the `pkgr plan` operation. I don't think we can get around this.
- Any hashing I'm doing is probably unnecessary, but it makes me feel a little safer knowing that we'll never accidentally try to install the wrong tarball just because something else was already in the cache with the same name. This -shouldn't- happen either way, since we redownload/re-unpack tarballs every time.

PR #249, opened 2020-03-31
Fix/linux env bug

Dreznel:
addresses #246


Dreznel:
Need to update unit tests, and I still wanted to try out what using Library instead of a temp folder would look like.

PR #260, opened 2020-04-28
Addressing Tarball-Read-Order bug

### LINKED ISSUE #259: Some tarballs can't be unpacked by pkgr
As it turns out, the method we use for unarchiving Tarballs (for "additional package" installation specifically) will sometimes fail. 

For some tarballs, the first item read from a gzip Stream is _not_ the root directory of the archive. That means that pkgr tries to unpack files into a directory that it hasn't bothered to create yet, thus causing an error. 

To fix this, we're going to strip out our custom unarchiving code and use https://github.com/mholt/archiver instead.

Dreznel:
fixes #259

Merging to release 1.0 for review purposes only. We haven't back-merged the hardened release into dev yet, but once that is done, we should merge this fix into dev and cut another release.


dpastoor:
adjusted after merging

PR #271, opened 2020-06-08
Add Linux binary support, RepoType, RepoSuffix to pkgr

evol262:
To support RSPM like repositories and binary builds of CRAN repos, we can make the following changes to pkgr:

Add a "RepoType" customization which allows CRAN (default), MPN, or RSPM

If it's an MPN repo and we're running on a supported version of Linux, try to get Linux binaries. To make this path configurable, there is an optional RepoSuffix field which can be used to customize the URI.

As an initial step to all of this, add a drone config and some basic tests for CI

In general, reviewing this makes it obvious that there are far too many similar/identical structs spread across packages with different initializations, making it difficult to determine which struct you are actually working with and makes changes unnecessarily complicated (cmd -> configlib -> cran -> repodb just to add these types, with multiple RepoConfig structs which aren't identical. Why not just have one which has everything we need and pass by reference instead of "RepoConfig | RepoDb | RepoURL", where some or all fields may be the same, but don't hold the same values, where we could instead have RepoDb nested under RepoConfig, and RepoURL as plain fields


cran/repodb.go
```diff
@@ -87,12 +92,15 @@ func (repoDb *RepoDb) Hash(rVersion string) string {
...
-func GetPackagesFileURL(r RepoURL, st SourceType, rv RVersion) string {
+func GetPackagesFileURL(r *RepoDb, st SourceType, rv RVersion) string {
 	if st == Source {
-		return fmt.Sprintf("%s/src/contrib/PACKAGES", strings.TrimSuffix(r.URL, "/"))
+		return fmt.Sprintf("%s/src/contrib/PACKAGES", strings.TrimSuffix(r.Repo.URL, "/"))
 		// TODO: fix so isn't hard coded to 3.5 binaries
 	}
-	return fmt.Sprintf("%s/bin/%s/contrib/%s/PACKAGES", strings.TrimSuffix(r.URL, "/"), cranBinaryURL(rv), rv.ToString())
+	if r.RepoSuffix != "" {
+		return fmt.Sprintf("%s/bin/%s/PACKAGES", strings.TrimSuffix(r.Repo.URL, "/"), r.RepoSuffix)
```
dpastoor, line 101: I don't think this is sufficient as a suffix - I see Suffix as truly being an additional component of the fully qualified "linux binary url" 

```
<uri>/bin/linux/<distro>/<reposuffix>/contrib/<rversion>/<pkg>_<version>_R_<platform>.tar.gz
```

In addition, if we do drop the linux/distro stuff, there still should be some differentiation around R versions to provide some consistency and make sure pkgr continues to operate in the multiple R version environment smoothly without having to manually set R Version numbers in the Suffix itself

For example, valid suffix examples might be:

```
20.03
metworx/20.06
```

```
/bin/linux/ubuntu/bionic/20.03/contrib/3.6
/bin/linux/ubuntu/bionic/20.03/contrib/4.0
/bin/linux/ubuntu/bionic/metworx/20.03/contrib/3.6
/bin/linux/ubuntu/bionic/metworx/20.06/contrib/4.0
```

cran/repodb.go
```diff
@@ -31,10 +31,15 @@ func NewRepoDb(url RepoURL, dst SourceType, rc RepoConfig, rv RVersion) (*RepoDb
 		repoDatabasePointer.DefaultSourceType = rc.DefaultSourceType
 	}
 
-	if SupportsCranBinary() {
+	if SupportsCranBinary() && rc.RepoType == MPN {
```
dpastoor, line 34: as discussed on slack - this causes mac/windows binaries to not acquire binaries - see #272 or #273 for alternative implementation(s)

cran/utils.go
```diff
@@ -87,3 +129,37 @@ func RepoURLHash(r RepoURL) string {
...
+	//os-release doesn't consistently quote variables, so we need to manipulate it a little bit
+	configData, err := ioutil.ReadFile("/etc/os-release")
+
+	//Find all unquoted strings and quote them
+	re := regexp.MustCompile(`(.*?=)([^"].*)`)
+	fixedConfig := re.ReplaceAll(configData, []byte("${1}\"${2}\""))
+
+	//Let viper map it in otherwise it defaults to yaml
+	//Don't pollute the global viper instance or we'll segfault later trying to read pkgr.yml
+	vp := viper.New()
```
dpastoor, line 147: I'm uncertain why the use of viper vs just using the toml unmarshaller directly? 

```
	re := regexp.MustCompile(`(.*?=)([^"].*)`)
	osRel, err := ioutil.ReadFile("os-release")
	if err != nil {
		panic(err)
	}
	fixedConfig := re.ReplaceAll(osRel, []byte("${1}\"${2}\""))
	osr := OsRelease{}
	err = toml.Unmarshal(fixedConfig, &osr)
	if err != nil {
		log.Fatal("could not unmarshall /etc/os-release with err: %v", err)
	}
```

cran/repodb.go
evol262: Let's go with #273. Conflict is resolved (and we keep the rversion and /contrib in another patch in this branch)

cran/utils.go
evol262: TOML unmarshaling is less flexible than Viper (which modifies it somewhat), and /etc/os-release fields are not consistent across distros, which is why there are some pointers (for nullable fields)

PR #272, opened 2020-06-11
fix: altering NewRepoDb function to use separate conditions for CRAN-…

Dreznel:
…binaries and Linux Binaries.

Also fixing a broken unit test.

## Problem
I ran into a bug where pkgr would not install binaries for R3.6 on MacOS (Mojave). After looking into it, I believe the changed in this PR is the culprit.

I have confirmed that the changes fix _my_ issue on Mac, but I have not yet re-tested the RepoType behavior.

Feel free to merge this PR or make your own version of the changes as you see fit!


Dreznel:
I have confirmed that RepoType: MPN still works after this code change.

PR #273, opened 2020-06-11
refactor: binary logic

dpastoor:
Simplification to map the logic around just generalized "does it support binaries" instead of CranBinary specifically


evol262:
Looks good to me. Let's go this way instead of the other

PR #295, opened 2020-09-21
fix: attempt ignore if Path set for custom package version

dpastoor:


dpastoor (commented):
merging this down - tests are incomplete to new changes however not fully fleshed anyway

PR #309, opened 2020-11-12
METWORX-202 Strip conda from the front of the path if present

evol262:
Since R uses its own compiler/linker configuration, we have the
option of modifying Renviron and Makeconf, but this is not a 100%
reliable solution for mitigating problems.

Conda ships more or less as a complete chroot including system
utilities, and its own version of many libraries. We should not
let Conda override this, so strip it from the path instead. For
configurability, add an option to let users preserve this
(possible that it will never be used, but it's there)

### DIFF
--- a/rcmd/configure.go
+++ b/rcmd/configure.go
@@ -55,6 +55,10 @@ func configureEnv(sysEnvVars []string, rs RSettings, pkg string) []string {
 				continue
 			}
 			if evs[0] == "PATH" {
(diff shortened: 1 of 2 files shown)


dpastoor:
Closing this out - it was determined that as long as conda was not first on the path, the installation would proceed fine and thus this workaround was not needed.

As a solution this felt too magical and tied to metworx's conda configuration as well, rather than a general solution.

PR #331, opened 2021-01-07
Feature/2.0.2 validation

Dreznel:


dpastoor (approved):
I will merge this for the code changes, but we need to clean up whats in the local db's for testing badly, the checked in files are binary compiled mac versions, which is not ideal. We've discussed a significant cleanup anyway so we'll live with it as is for now

dpastoor:
actually i'm not sure why this was pointing to helper/large-file-commit,and why that isn't associated with a PR @Dreznel ?

Dreznel:
My apologies,

When we were validating in January, I made a branch to make tweaks and such to the automated & integration tests, but that ended up with me having to add packages that were compatible with R4.0. I made the "helper" branch to hold those changes separately just so I could easily see the diffs (at first) as I continued making changes to the testing files. Looks like this branch ended up growing a bunch too, but splitting them out was useful at the time.

It was my intent that this branch be pointed towards develop before merging, at which point it would have already included all of the changes from helper/large-file-commit anyway. That's why you didn't see a PR for the "helper" branch.

As for why we're using binaries, in some cases that's because I wanted to test things related to pkgr's default behavior/pkgr's ability to install binaries specifically. In other cases, it's probably just an oversight, or the result of taking the path of least resistance for getting things runnable locally for me.

dpastoor:
ok lets merge this down then - can you create a pr from this branch into develop and i'll pull it in. thanks for the context

PR #343, opened 2021-02-16
exp: allow for ignoring certain packages in the solve

dpastoor:
this will currently rip them out of the dep tree, so an ignored package
and all its deps will be removed. This means if that package is part of
the required package set it will cause the solve to fail, however
when having some suggested deps that want to selectively exclude
can work well.


dpastoor:
have been using this for MPN builds for months with no issues. Will tentatively merge but will need additional testing before release

PR #352, opened 2021-05-30
fix customization code

dpastoor:
One of the more subtle bugs that has emerged is due to the fact that viper is a bit loose around how it unmarshals, when combined with some interface{} related flexible code meant that when needing a structure that actually mapped to the actual expected config more explicitly, some failures occured.

In particular, with the add/remove commands, explicit yaml unmarshalling vs viper was used and was getting parsing errors in the yaml. This actually showed that the exisitng underlying struct vs the user specification was off. This has now been fixed, but also showed where a different underlying structure has been used throughout the code.

Namely, as a user customizations are specified as arrays:

```
Packages:
- dplyr:
    Suggests: true
- ggplot2:
    Repo: something
```

This structure is a slice of map[string], however within the code where just using a map structure directly. 

The code changes here fix the underlying structures to the user spec and fix the resulting code.

### DIFF
--- a/cmd/install_test.go
+++ b/cmd/install_test.go
@@ -68,10 +68,11 @@ func InitGlobalConfig(libraryPath, localRepo string, update, suggests bool, inst
 		//Logging: configlib.LogConfig{Level: "debug"}, // Controlled before cfg unmarshalling, I think
 		Cache: "./testsite/working/localcache",
 		Customizations: configlib.Customizations{
-			Repos: map[string]configlib.RepoConfig{
+			Repos: []map[string]configlib.RepoConfig{{
 				"testRepo": configlib.RepoConfig{
 					Type: installType,
 				},
+			},
 			},
 		},
 		//LibPaths: nil,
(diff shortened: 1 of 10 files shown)


dpastoor:
The only user facing change of this structure now is actually pkgr add works properly when customizations are present (it would work without customizations fine before since if no customizations present didn't need to unmarshall to improper yaml, but when customizations present would error)

before this fix:

given a pkgr.yml

```
Version: 1
Threads: 6
Packages:
  - R6
Repos:
  - MPN: https://mpn.metworx.com/snapshots/stable/2021-06-20
Lockfile:
    Type: renv
Customizations:
  Packages:
  - R6:
      Type: source
```

and run `pkgr add dplyr`

```
FATA[0000] yaml: unmarshal errors:
  line 23: cannot unmarshal !!seq into map[string]configlib.PkgConfig
```

After this succeeds. Going to merge this branch in as we'll be combining a bunch of new functionality then refactoring the entire test suite with some of the patterns introduced #360

PR #366, opened 2021-07-20
rollback testing

dpastoor:
This PR implements testing rollback.

## userland changes

I also changed the rollback flag structure, the previous version was confusing compared to the other negative assertions we have like `no-secure`, so this normalized so also have `--no-rollback` instead of `--rollback=false`

## testing

This, as mentioned in internal discussions, is one that could theoretically be difficult to test, however with the new patterns was not so bad.

In particular, instead of needing to construct a fake environment, I simply leveraged the fact we can build a partial environment in the first install, then subsequently do additional installs to try the rollback.

One thing I don't have a good feel for and am looking for feedback is some of how to construct the Run commands and some of the "setup-y" code - eg building the initial command. Should it go in the Run, outside the run, contextually different choices?


dpastoor:
One thing with CI i'm somewhat confused on as is to why it keeps re-downloading test dependencies when running the tests each time. I'm not sure where they're being dropped into and why after even the first test they need to be reinstalled again over and over.

elisarver (approved):

cmd/install_test.go
```diff
@@ -60,7 +60,7 @@ func InitGlobalConfig(libraryPath, localRepo string, update, suggests bool, inst
 	cfg = configlib.PkgrConfig{
 		Threads:  5,
 		Update:   update,
-		Rollback: false,
+		NoRollback: false,
```
Dreznel, line 63: This is now doing the opposite -- is this what we want? I assume we're scrapping these tests anyway, but this could technically cause problems.

Dreznel: GitHub won't let me comment on it as part of the review, but on line 296, we're still setting the "Rollback" field of cfg.

configlib/structs.go
```diff
@@ -48,7 +48,7 @@ type PkgrConfig struct {
 	Suggests       bool                `yaml:"Suggests,omitempty"`
 	NoRecommended  bool                `yaml:"NoRecommended",omitempty"`
 	Repos          []map[string]string `yaml:"Repos,omitempty"`
-	Rollback       bool                `yaml:"Rollback,omitempty"`
+	NoRollback     bool                `yaml:"no_rollback,omitempty" mapstructure:"no_rollback,omitempty"`
```
Dreznel, line 51: The underscore doesn't follow the convention we've established for things like "NoSecure" and "NoRecommended"

integration_tests/rollback/Rscripts/Rscripts.Rproj
```diff
@@ -0,0 +1,13 @@
+Version: 1.0
```
Dreznel, line 1: Should this file be .gitignored?

integration_tests/rollback/rollback_test.go
```diff
@@ -0,0 +1,82 @@
...
+		t.Run("in configuration file", func(t *testing.T) {
+			res, err := installCmd.Run(ctx, "pkgr", "install", "--config=pkgr-no-rollback-tarball.yml", "--logjson")
+			if err != nil {
+				t.Fatalf("could not install baseline packages with err: %s", err)
+			}
+			rollbackOutputCheckHelper(t, res, RollbackDisabledInstalled)
+		})
+
+		t.Run("as CLI flag", func(t *testing.T) {
+		    err := os.RemoveAll("test-library/ps")
```
Dreznel, line 55: This feels a bit unstable to me. In general, I think it's better if each test is sort of working in a "fresh" environment and doesn't need bits of cleanup like this. Otherwise, we risk breaking this test when we change the order in which things happen.

I would instead consider adding a pkgr.yml file with a different test-library for tests like this. YMMV.

Dreznel (changes requested):
My only major gripe is that the `no_secure` with an underscore doesn't follow the previously established convention for multi-word arguments in the pkgr.yml file. 

Everything else is just something for consideration.

cmd/install_test.go
dpastoor: you're right - we need to swap these - though I'm wondering if we even need these tests anymore - eg prob can/should just leverage the integration tests

configlib/structs.go
dpastoor: so we're in a "bad" spot around naming - newer stuff has used snakecase, older stuff camelcase. I stared at it a while myself between NoRollback and just went crap. One thing we should potentially look at is aliases - eg just allow either to be used.

integration_tests/rollback/Rscripts/Rscripts.Rproj
dpastoor: nope - don't .gitignore Rproj since that way can easily launch in rstudio - which will properly give you the wd context etc.

Basically, between this + the makefile the entire test suite is set up to interactively use and poke very easily

integration_tests/rollback/rollback_test.go
dpastoor: i see where you're coming from, also debated that myself, the issue that popped up was basically... we have a ton of these combinations, and reconstructing the entire environment each time is going to be tedious. I was hoping to have some middle ground of cleanup when doing the same test multiple times, at least when the "cleanup" is simple. Now that I think about it though, this wouldn't actually need to completely re-fetch packages since it would actually pull from the previous steps install cache if present, but also if run in different order would still work and just do the entire pull. I think we can change this.

dpastoor: @Dreznel see the commit below for an update - now i am resetting the entire library

PR #367, opened 2021-07-21
Test/outdated

Dreznel:
Learning the new testing pattern and applying it to the outdated packages integration test.

I branched off of Devin's `test/outdated` branch, which is why I'm merging it back into that. We can just point this to `exp/testing` or `develop` if we need to.


Dreznel:
I kept all four of the "outdated" packages because I believe they were testing different scenarios:

* R6: Basic update with no dependencies
* crayon: Not listed in pkgr.yml, but is a dependency of pillar, which IS listed in pkgr.yml. Also wasn't installed by pkgr. 
* pillar: Basic update with dependencies (e.g. crayon)
* Matrix: I have a note that this was kept in as a regression test for some older issue. We can probably remove this.

Dreznel:
Could be cleaner, but in the interest of getting through these tests I'm submitting for review now.

integration_tests/outdated-pkgs/pkgr-update.yml
```diff
@@ -0,0 +1,16 @@
+Version: 1
+# top level packages
+Packages:
+  - R6
+  - pillar
+  - Matrix
```
dpastoor, line 6: this is a heavy package IIRC and takes a bit of compile time to install, probably better to choose something simpler. What was the reason to add Matrix into the mix?

Dreznel: Yeah we can probably remove Matrix, see my comment here: https://github.com/metrumresearchgroup/pkgr/pull/367#issuecomment-884409620

PR #368, opened 2021-07-22
Test/load

Dreznel:
Automating the "load" and "load-fail" E2E tests.


Dreznel:
I set this one to merge to exp/testing because that seems to be the parent branch of everything. I assume that the previous two pull requests (test/outdated and test/rollback) should be merged before this one.

PR #374, opened 2021-07-25
ci: adjust go env so can use mod cache to not re-dl deps for tests

dpastoor:


dpastoor:
evidence in CI:

baseline will still pull all the test deps:

image

but subsequent steps don't download goldie/testify/etc

image

dpastoor:
getting this into develop so others can use it too

PR #375, opened 2021-07-26
pull request to track automation of E2E tests

Dreznel:


Dreznel:
Issue #376 is going to cause tests to fail for a bit.

.drone.yml
```diff
@@ -161,7 +161,7 @@ steps:
       - cd integration_tests/library
       - make test
   - name: mixed-source-and-customizations
-    image: 906087756158.dkr.ecr.us-east-1.amazonaws.com/r-dev-ci:4.0.0
```
dpastoor: If we need an R 4.0 image its just R 4.0.5, not 4.0.0

Dreznel: I just tried this and it still didn't work, unfortunately.

PR #377, opened 2021-08-01
refactor: use noupdate to normalize to the inverse pattern elsewhere and change default

dpastoor:
it has found that the default behavior to not update is generally not what users
want, especially around reproducibility. Instead, we want users to be
able to not update only if they absolutely need to.

In addition, part of this original decision was to prevent packages attempting to be
updated causing major failures that couldn't get back to original state.
This is no longer an issue given the rollback functionality.


Dreznel:
@dpastoor I think I found a fairly major problem with the change:
image

The above screenshot shows that the multi-repo test is failing after the change. I believe this is because the --update flag used to scan _all_ listed repositories for later versions of packages. Since that is now the default, **pkgr will only respect repository order if the NoUpdate flag is false.**

PR #378, opened 2021-08-04
tidying up some tests to make it easier to run them consistently…

Dreznel:
… alongside the unit tests.

The refactoring I did in the outdated tests ended up being a red-herring (the problem was that pkgr wasn't rebuilding before running tests), but I still think the changes are valid. I just don't want to muddy the existing code review, so I'm opening this as a separate PR.


Dreznel:
One thing to note is that I've added `LC_ALL=en_US.UTF-8` into many of the `.Renviron` files. This addresses the discrepancies we are experiencing when running Golden-File tests through Goland versus the Mac Terminal.

Dreznel:
One issue I'm running into is this:

image


The `clean pkgdbs` command is hard to test consistently. Since the actual location of the pkgdb varies by OS, I'm currently extracting the location of the the pkgdb cache from the pkgr logs, then using that in my assertions. This is certainly problematic in itself, but this at least verifies that pkgr is using what it claims to be using as far as caching goes. 

However, since I cannot reliably predict where the cache will be prior to running it, I can't "clean" it programmatically at the beginning of the test. This means that, when I run the `clean pkgdbs` test alongside other tests, the global cache gets muddied, and the `clean pkgdbs` test fails falsely.

Dreznel:
Okay, I think I have figured out the source of the weird test-inconsistencies. Shout out to this article: https://medium.com/@xcoulon/how-to-avoid-parallel-execution-of-tests-in-golang-763d32d88eec

Basically, go tests from different packages run in parallel. I knew that parallel testing would be a problem, because one of the tests deletes the global `pkgdb`, which would obviously cause problems to other in-progress tests. However, since I wasn't using `t.Parallel` anywhere, I thought that my tests would run "single-file" by default. As the article describes, I need to ad `-p 1` to my `go test` commands.

Additionally, I believe that this resource collision got my tests in an invalid state overall, which caused the Tarball Installation tests to fail. This _might_ account for the `outdated-pkgs` issue described in #377 , but I am uncertain so far.

Dreznel:
All remaining automated tests pass when run with the `-p 1` flag. 

image

PR #385, opened 2021-09-15
Add go.uber.org/automaxprocs to cmd/pkgr/pkgr.go

elisarver:
This one is simple. As noted earlier in discussion, we should limit our application so it doesn't run out of allotted time-slice and respects system limits.


elisarver:
The merge is blocked because of a test `mixed-source-and-customization` that is failing in develop as well.

elisarver:
The failing test is caused by `#386`

dpastoor:
Thanks Eli - great to get this in quick.

Seems pretty reasonable - I think it will be worth doing some quick sanity checks, in particular I assume this change should not impact the following:

* explicitly setting threads should still win out
* no change on mac/windows/linux when not in a container/quota'd environment

what should change is if we are. This to me seems something that would be more difficult than likely worth to build a long term automated test, rather some evidence once that its working feels sufficient to me.

Also in looking briefly at the uber lib, looks like this can also be overridden by an environment variable, but they reference it as a variable like _gomaxprocskey or the like - I didn't chase that down. Would be good to doc that there is that control functionality as well (what that key specifically is).

elisarver:
>     * explicitly setting threads should still win out

This remains unchanged:

```go
// If user has not specified a thread count themselves, will limit the user to 8 threads max to avoid issues.
func getWorkerCount(threadCount, numCpus int) int {
	var nworkers int
	if threadCount < 1 { // This indicates that the user has not specified a thread count, i.e. we're using the default thread count of "0".

		workerCap := 8 // We have decided to cap this number at 8 unless the user requests otherwise. This is to prevent issues with cyclic Suggests installs.
		if numCpus <= workerCap {
			nworkers = numCpus
		} else {
			nworkers = workerCap
		}
	} else {
		nworkers = threadCount
		if nworkers > numCpus+2 {
			log.Warn("number of workers exceeds the number of threads on machine by at least 2, this may result in degraded performance")
		}
	}
	return nworkers
}
```

With `threadcount = 12` and numCPUs at 2, this will warn and return 12.

>     * no change on mac/windows/linux when not in a container/quota'd environment

There are two files in the Uber library:

cpu_quota_linux.go
cpu_quota_unsupported.go

The first is tagged for linux, the second for !linux. The !linux case returns `-1, QuotaUndefined, nil` unconditionally.

If you have GOMAXPROCS set in the environment, it uses that, as well.

So you still have both methods of control to run with with lots of workers/threads.

elisarver:
If I set GOMAXPROCS=14 (my laptop has 12) I still get:

> WARN[0003] number of workers exceeds the number of threads on machine by at least 2, this may result in degraded performance

elisarver:
I also had to change plan to use `GOMAXPROCS(0)` in stead of `NumCPU()`.

```
$ export GOMAXPROCS=6
$ ../../cmd/pkgr/pkgr plan --threads 14
…
INFO[0000] Installation would launch 14 workers  
…
$ ../../cmd/pkgr/pkgr plan
…
INFO[0000] Installation would launch 6 workers
...
$ ../../cmd/pkgr/pkgr install --threads 14
…
WARN[0000] number of workers exceeds the number of threads on machine by at least 2, this may result in degraded performance 
…
```

elisarver:
Also, using `NumCPU()` caused a bug where when `NumCPU > GOMAXPROCS`, `GOMAXPROCS` was being ignored/lost.

elisarver:
@dpastoor you have any thoughts on the response?

dpastoor:
Fantastic, lgtm!

Issue #387, opened 2021-09-23
upgrade version of command


elisarver:
command@v0.1.0 is now out, update tests to use it.

Previously the command.New accepted functional options, but the exec.Cmd rework of the package has made this more go-stdlib-like.

before: 

```go
ctx := context.TODO()
cmd := command.New(command.WithDir("somewhere"))
output, err := cmd.Run(ctx, "echo", "hello world")
```

after:

```go
cmd := command.New("echo", "hello world")
cmd.Dir = "somewhere"
output, err := cmd.CombinedOutput()
```

PR #388, opened 2021-09-23
Update command package

elisarver:
This follows #387 and updates the tests to use the new library format.


dpastoor:
Cool deal! I can take a look tomorrow afternoon

dpastoor (approved):
LGTM - does make me slightly wonder about the whole "should we return bytes or just ack that most people want a string". Still feels right to do bytes.

Failing test is expected to still be failing

elisarver:
> LGTM - does make me slightly wonder about the whole "should we return bytes or just ack that most people want a string". Still feels right to do bytes.

There's a pretty strong reason the standard lib is all-in on `[]byte` rather than strings: `[]byte` are mutable, and addressable by cell. And `string` are immutable. Lots of allocation when you handle strings.

elisarver:
I'm pretty confident in my results here, since it's a simple transform. Merging.

PR #390, opened 2022-01-04
Update go.mod to patch around a dependabot vulnerability.

elisarver:
Update the dependency of xz indirectly


seth127 (approved):
Looks good. [Initial alert](https://github.com/metrumresearchgroup/pkgr/security/dependabot/go.sum/github.com%2Fulikunitz%2Fxz/open) says moving to `xz >= 0.5.8` fixes the issue and I can see this update moves us to `xz 0.5.9`

PR #393, opened 2022-01-13
Fix illegal use of .ArtifactName in goreleaser

elisarver:


seth127 (approved):
looks good. 🤞 it works in drone.

PR #399, opened 2022-02-21
remove extra space in log message

billdenney:
There is what appears to be an extra space in the log message for "downloading package".  If it's intentional, no worries about closing this.


kyleam (approved):
Thanks for spotting that.

It doesn't look like [any tests](https://github-drone.metrumrg.com/metrumresearchgroup/pkgr/212) are thrown off by this change, and I think it's unlikely that any consumers of the JSON logs were relying on the trailing space here.

PR #408, opened 2023-06-05
config: mark renv library path for more reliable detection

kyleam:
The last commit of fixes renv detection when other code startup code writes to stdout.  I hit into a case in the wild where someone's .Rprofile `cat` some text, leading to `pkgr` aborting because it received more than one line on stdout.  (Cases that use `message` don't hit into this because that writes to stderr.)

Demo:

```
$ head -1 .Rprofile
cat("bah\n")

$ pkgr plan
WARN[0000] expected one line for renv library, got 2:
 ["bah" "/data/home/kylem/src/github/metrumresearchgroup/nmrec/renv/library/R-4.1/x86_64-pc-linux-gnu"]
FATA[0000] Lockfile type is renv, but calling renv to find library path failed.
Check that renv is installed or consider using `Library:` instead.
```

The first two patches clean up an unrelated test failure that I noticed when pushing this branch.  It looks like it's due to a dead https://cran.microsoft.com link.

### DIFF
--- a/configlib/config.go
+++ b/configlib/config.go
@@ -9,6 +9,7 @@ import (
 	"os/exec"
 	"path/filepath"
 	"runtime"
+	"strings"
 
 	homedir "github.com/mitchellh/go-homedir"
 	log "github.com/sirupsen/logrus"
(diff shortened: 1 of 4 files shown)


kyleam:
Hmm, the drone runs aren't showing up, but they were triggered and passed:

 * https://github-drone.metrumrg.com/metrumresearchgroup/pkgr/240
 * https://github-drone.metrumrg.com/metrumresearchgroup/pkgr/241

kyleam:
Thanks for the review @seth127.  I noticed an unintended behavior changed in the bad_customization_test.go test fix, so I'll push an update for that.

kyleam:
Passing drone runs for the update (079ab93):

 * https://github-drone.metrumrg.com/metrumresearchgroup/pkgr/244
 * https://github-drone.metrumrg.com/metrumresearchgroup/pkgr/245

PR #420, opened 2024-08-14
Delete some unreachable code

kyleam:
This series removes some files/packages that aren't currently connected to the pkgr executable or used in any other way.  These were flagged by the per file coverage.


seth127 (approved):
This all looks like sensible clean up, and the explanations in the commit messages make sense to me. Thanks for documenting.

PR #427, opened 2025-04-01
CI maintenance updates

kyleam:
* Switch the OS to Ubuntu 22.04 to prepare for Ubuntu 20.04 runners being retired this month

   * Adjust tests for compatibility with Ubuntu 22.04

 * Update a test assertion to account for a change in renv 1.0.6

 * Update R versions to match the two most recent R versions on the latest Metworx

 * Update the Go version for the check jobs to match what ships with the latest Metworx


cmd/install_test.go
```diff
@@ -202,25 +202,22 @@ func TestTarballInstall(t *testing.T) {
 			installUpdates:  false,
 			installSuggests: false,
 			toInstall: []string{
-				"pillar",
+				"openssl",
 			},
 			toInstallTarballs: []string{
-				filepath.Join(localReposDir, "tarballs", "testthat_2.1.1.tar.gz"),
+				filepath.Join(localReposDir, "tarballs", "pillar_1.3.1.tar.gz"),
```
seth127, line 208: Good change here. I appreciated the detailed explanation in the commit message.

kyleam:
@seth127 Thanks for reviewing.

PR #432, opened 2025-11-18
Update tests for compatibility with R 4.5

kyleam:
A couple of tests are failing under R 4.5 because they use a version of digest that doesn't build on R 4.5.  Update each spot to use an MPN snapshot that has a digest version that's compatible with R 4.5.


seth127 (approved):
Looks good. Thanks for the updates.

